PR #24, opened 2018-12-06
Use correct NZIC link in menu

puqeko:


puqeko:
Added to deployed branch in 886480b7745ee533e26446a6467750256a1d3828.

PR #28, opened 2019-02-16
Fixes for install scripts, update to Ruby 2.3.8

tom93:
Various fixes for the install scripts. Update to Ruby 2.3.8 because earlier versions have reached end-of-life and custom evaluators only work on Ruby >= 2.3. The tests now pass on Travis CI.

 - 8a11cce Change to openjdk-8 as 6 is not supported any more
 - 54a82ca Fix isolate root path in debootstrap script
 - f54dceb Comment out useless 'apt-get update' in debootstrap script
 - d1451da Install software-properties-common before using add-apt-repository in debootstrap script
 - 9855aba Create script/update.time in install.bash (like update.bash)
 - b07994e Warn about using Ruby < 2.3
 - d6b301e Update Ruby from 2.0.0 to 2.3.8
 - e343d3c Import RVM's GPG signing keys before installing
 - 036456a Only install 'bundler' system-wide if 'gem' was installed system-wide
 - 2561ac8 Don't use sudo with bundler
 - 8c43255 Detect bundle errors in install scripts that use pipes '|'
 - e2e1067 Fix 'Unknown key: null' in create_settings db migration
 - 44caf13 Workaround for missing test/unit/assertions when running specs
 - a191d94 Update qless from 0.10.3 to 0.12.0


logan12358:
Just letting you know, this all looks good but I'm not sure 100% where best to merge it to. Currently the server is running off the [`deployed`](https://github.com/NZOI/nztrain/tree/deployed) branch. It might make things confusing to put some of this there, because we wouldn't actually be re-deploying with the new scripts. The `master` branch should probably converge with `deployed`, but I haven't looked into that. Do you have thoughts on what best to do?

tom93:
Thanks for taking a look. I wasn't sure which branch to work on and picked `deployed` because it had more activity. It makes more sense to develop on branch `master` and merge the changes into `deployed` when the server is updated.

**Convergence**

The branches `master` and `deployed` have diverged:
 - `master` contains pull request #23 ("Fixes") with three commits
 - `deployed` contains a large commit with mostly the same changes (0b560f9 "Changes that Andrew made on the server", [diff](/NZOI/nztrain/compare/cde6324..0b560f9)) and puqeko's recent work

Changes that are only in `master`:
 - update Ruby in .travis.yaml from 2.0.0 to 2.2.9 (c8e987f, part of #23)

Changes that are only in `deployed`:
 - update Ubuntu release of chroot from "precise" to "xenial" (part of 0b560f9)
 - update package name of GHC from "ghc6" to "ghc" (part of 0b560f9)
 - puqeko's recent work

I think merging them is a good idea. If it's done now (before merging my pull request) there won't be any conflicts. I suggest that `deployed` is first merged into `master` (to present `master` as the mainline branch) and then `master` is merged into `deployed` using a fast-forward merge (to make it clear they are both at the same state).

The target branch of my pull request can then be changed from `deployed` to `master` (there will only be a trivial conflict in .travis.yaml).

**Inconsistency with deployment**

I think the inconsistency in updating the scripts without re-deploying is unavoidable: the updated scripts are needed for new installations so they should be in the development branch (otherwise it's hard to develop), and the development branch needs to (eventually) be merged into the server's branch. My suggestion is to go over the changes and apply them to the server manually if desired. Let's discuss that over email. If anyone else is also interested, let me know and I'll CC you.

lfairy:
@tom93 you should have push access to this repo now, courtesy of Logan. Would you like to do the merge?

lfairy:
Thanks Tom!

PR #38, opened 2019-06-17
Fixes for front end, su, COCI importer, specs

tom93:


Gemfile
```diff
@@ -2,7 +2,7 @@ source 'http://rubygems.org'
 
 gem 'rails', '~> 4.0.0'
 
-gem 'nzic_models', github: 'NZOI/nzic_models'
+gem 'nzic_models', git: 'https://github.com/NZOI/nzic_models.git'
```
lfairy: Can we upgrade to Bundler 2.0 instead, or at least enable the HTTPS option?

See bundler/bundler#4978

app/assets/stylesheets/submission.css
```diff
@@ -27,63 +27,44 @@ table.results {
...
-  background-color: #E03030; /* red */
-  border: 2px solid #681818;
-}
-
-.results tr.test_case.status_correct {
-  background-color: #60C040; /* green */
-}
-
-.results tr.test_case.status_pending {
-  background-color: #6060FF; /* blue */
```
lfairy: Nice simplification!

app/workers/problem_series/import_worker.rb
```diff
@@ -49,7 +49,10 @@ def perform
               problem.problem_sets << problem_set
 
               # set problem weighting in official problem set
-              ProblemSetProblem.find_by(problem_set_id: problem_set.id, problem_id: problem.id).update_attributes(weighting: manager.get_maxscore)
+              # can't use problem.problem_set_associations.find_by(problem_set_id: ...) because it
+              # operates on the database rather than the cache, and the problem might not have been
+              # saved yet
+              problem.problem_set_associations.select{|assoc|assoc.problem_set_id == problem_set.id}.first.update_attributes(weighting: manager.get_maxscore)
```
lfairy: I'm not familiar with Rails, but I wonder if there is a more idiomatic way to express this.

Also, do we want to add some spaces in the closure?

```suggestion
              problem.problem_set_associations.select{ |assoc| assoc.problem_set_id == problem_set.id }.first.update_attributes(weighting: manager.get_maxscore)
```

tom93:
Thanks for reviewing.

Gemfile
tom93: > Can we upgrade to Bundler 2.0 instead, or at least enable the HTTPS option?

We should upgrade to Bundler 2.0 eventually, but it's non-trivial because Rails 4.x requires Bundler < 2.0.

The github shortcut will be removed anyway at some point[1], so if we want to use it we should just add the directive

    git_source(:github) { |repo| "https://github.com/#{repo}.git" }

like in the newer Rails and Bundler templates[2][3]. I pushed a commit that does that (dccb9a4).

[1] <a href="https&#58;//&#103;ithub.com/bundler/bundler/issues/5699"><span>bun</span>dler#5699</a> (originally [intended to be removed in 2.0](https://github.com/bundler/bundler/pull/5705/commits/598ab4bef8fefd0a11d8b4b735a6c13b2cfb2e78#diff-ea5f81b03d4209ef02c02845cd95fcbfR461), then [delayed to 3.0](https://github.com/bundler/bundler/pull/7000/commits/8cbbb3985607bbfffa2d26666da163631c4c5acd), and then [delayed further](https://github.com/bundler/bundler/pull/7000/commits/22db310f67c717a972cd5437cf6dcfc4bfc76b26))  
[2] [rails/railties/lib/rails/generators/rails/app/templates/Gemfile.tt](https://github.com/rails/rails/blob/60e19c6d6cb60159af4e2247f29d2cea375a21d6/railties/lib/rails/generators/rails/app/templates/Gemfile.tt)  
[3] [bundler/lib/bundler/templates/Gemfile](https://github.com/bundler/bundler/blob/391b11966efb140a9ba204ca4e05ec7694296ce0/lib/bundler/templates/Gemfile)

app/workers/problem_series/import_worker.rb
tom93: > I'm not familiar with Rails, but I wonder if there is a more idiomatic way to express this.

Improvements welcome, they can be done as separate commits.

> Also, do we want to add some spaces in the closure?

I added spaces in the new code (commit 48d2345), but not the rest of the file (e.g. [line 214](https://github.com/NZOI/nztrain/blob/76733803893c5e7d2565230fc2c950f01e3c79dd/app/workers/problem_series/import_worker.rb#L214)).

lfairy:
No worries ☺️

What's the process for getting this deployed? @logan12358

PR #39, opened 2019-09-15
Add C++17

serverhiccups:
This should hopefully be a quick fix that allows submissions to be written in C++17. It is very possible that I missed a change that I needed to make or this breaks something so please check this code yourself (I can't run this on my mac).

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -54,7 +54,8 @@ class Language < ActiveRecord::Base
   def self.submission_options
     latest = LanguageGroup.where(identifier: %w[c++ c python haskell java ruby j]).pluck(:current_language_id)
     old = Language.where(identifier: %w[c++03 python2]).pluck(:id)
-    languages = Language.where(:id => latest).order(:identifier) + Language.where(:id => old).order(:identifier)
+    experimental = Language.where(identifier: %w[c++14]).pluck(:id)
+    languages = Language.where(:id => latest).order(:identifier) + Language.where(:id => old + experimental).order(:identifier)
     Hash[languages.map{ |language| ["#{language.group.name} (#{language.name})", language.id] }]
   end
 
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -54,7 +54,7 @@ class Language < ActiveRecord::Base
   def self.submission_options
     latest = LanguageGroup.where(identifier: %w[c++ c python haskell java ruby j]).pluck(:current_language_id)
     old = Language.where(identifier: %w[c++03 python2]).pluck(:id)
-    experimental = Language.where(identifier: %w[c++14]).pluck(:id)
+    experimental = Language.where(identifier: %w[c++14 c++17]).pluck(:id)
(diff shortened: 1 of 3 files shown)


serverhiccups:
I can see that it failed CI but I don't think that issue is related to my pr.

puqeko:
Thanks. It might not be that simple since we probably have an old GCC version. @logan12358 could you please check this?

Also, there is the question of whether we should add it or not. Currently, the IOI only go up to gnu++14 [https://ioi2019.az/en-content-26.html](https://ioi2019.az/en-content-26.html).

tom93:
I don't have a strong opinion on whether we should add it or not. For what it's worth the IOI 2019 page says they were considering adding C++17 as an option.

---

The current version of C++ is still marked as c++11 in [db/languages.yml:10](https://github.com/NZOI/nztrain/blob/83163a5ff0a5cd22f6a1341dcb042c3efbd152ce/db/languages.yml#L10). The submission page only shows the current version of each language (and some hard-coded old versions, see the `submission_options` function in [app/models/languages.rb](https://github.com/NZOI/nztrain/blob/83163a5ff0a5cd22f6a1341dcb042c3efbd152ce/app/models/language.rb#L54)). We would need to mark the current version of C++ as c++17, or hard-code C++17 into `submission_options` (maybe as an "experimental" language).

We would need to update GCC. The server currently has GCC 4.9 ([script/install/debootstrap.sh](https://github.com/NZOI/nztrain/blob/83163a5ff0a5cd22f6a1341dcb042c3efbd152ce/script/install/debootstrap.bash#L158) specifically installs that version) and it doesn't support "-std=gnu++17".

serverhiccups:
For what it's worth, we could consider adding C++14 as the main option and C++17 as 'experimental', that way, when C++17 becomes the default (as I imagine it will eventually) we should only have to flip a switch.

serverhiccups:
Currently C++11 is still the default, and C++14 and C++17 are added like legacy languages. I think that we should consider setting C++14 as the default and leaving C++17 as experiment.

tom93:
I moved adding C++14 into a separate pull request (#42) since it is easier to deploy than C++17, and cleaned up the commit history of this pull request.

We will need to manually update the server before deploying this:

```
chroot ...
apt-get update
apt-get install gcc-9 g++-9
update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 75
update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
update-alternatives --remove gcc /usr/bin/gcc-4.9
update-alternatives --remove g++ /usr/bin/g++-4.9
apt-get remove gcc-4.9 g++-4.9
apt-get autoremove
```

P.S. Some tips for using Git:
 - use `git log --graph --decorate  --all` to visualize the commit history
 - use `git fetch` instead of `git pull` to download new commits without automatically merging
 - careful use of `git rebase` and force-pushing can produce clean histories (this should be avoided with branches that have been merged into other branches, but it's OK to do this with pull requests that haven't been merged yet)

puqeko:
See #45.

PR #42, opened 2019-11-17
Add C++14 as an experimental language

tom93:
Not marking it as the current version yet so we can test it before
making it the default.


puqeko:
See #45.

PR #45, opened 2020-01-18
Update GCC

puqeko:
Replacing #39 and #42 since we can't yet merge changes in master to the deployed branch.

Update GCC from 4 to 9 and add

- GNU++14 (C++14)
- GNU++17 (C++17)
- GNU11 (C11)

Deprecate

- GNU++03 (C++03)
- GNU99 (C99)

These languages will no longer be available to select from, but the server will still support their compilation so that legacy code may be rejudged and we don't have to remove things from the database.

Install process:
_Originally posted by @tom93 in https://github.com/NZOI/nztrain/pull/39#issuecomment-554689957_

```
chroot ...
apt-get update
apt-get install gcc-9 g++-9
update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 75
update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
update-alternatives --remove gcc /usr/bin/gcc-4.9
update-alternatives --remove g++ /usr/bin/g++-4.9
apt-get remove gcc-4.9 g++-4.9
apt-get autoremove
```

### DIFF
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -7,22 +7,28 @@ c++:
   :lexer: c++
   :compiled: yes
   :interpreted: no
-  :current: c++11
+  :current: c++17
   :extension: .cpp
   :source_filename: program
   :exe_extension: .exe
   :processes: 1
   :variants:
-    c++03:
-      :name: "C++03"
+    c++03:  # depreciated
+      :name: "C++03 (GCC 9)"
       :compiler: /usr/bin/g++
       :compiler_command: "%{compiler} -std=gnu++03 -O2 -o %{output} %{source} -lm"
-#      upgrade: c++11 # submissions will be upgraded if possible, if support is removed, it will be force-upgraded
-#      note: probably will never happen for this particular example... (or at least not for the next 10 years)
     c++11:
-      :name: "C++11"
+      :name: "C++11 (GCC 9)"
       :compiler: /usr/bin/g++
       :compiler_command: "%{compiler} -std=gnu++11 -O2 -o %{output} %{source} -lm"
+    c++14:  # Changes since C++11: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1319r0.html
+      :name: "C++14 (GCC 9)"
+      :compiler: /usr/bin/g++
+      :compiler_command: "%{compiler} -std=gnu++14 -O2 -o %{output} %{source} -lm"
+    c++17:  # Changes since C++14: https://isocpp.org/files/papers/p0636r0.html
+      :name: "C++17 (GCC 9)"
+      :compiler: /usr/bin/g++
+      :compiler_command: "%{compiler} -std=gnu++17 -O2 -o %{output} %{source} -lm"
 c:
   :name: C
   :lexer: c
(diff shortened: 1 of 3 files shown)


db/languages.yml
```diff
@@ -7,22 +7,28 @@ c++:
...
-  :current: c++11
+  :current: c++17
   :extension: .cpp
   :source_filename: program
   :exe_extension: .exe
   :processes: 1
   :variants:
-    c++03:
-      :name: "C++03"
+    c++03:  # depreciated
```
lambda-fairy: "Depreciate" is an accounting term. The word you're looking for is "deprecate" with no "i".

```suggestion
    c++03:  # deprecated
```

db/languages.yml
```diff
@@ -34,10 +40,15 @@ c:
   :exe_extension: .exe
   :processes: 1
   :variants:
-    c99:
-      :name: C99
+    c99: # depreciated
```
lambda-fairy:
```suggestion
    c99: # deprecated
```

tom93:
Is there a reason to include C++14 rather than skip directly to C++17?
I only created #42 (Add C++14) as an intermediate step until we could update GCC and add C++17. Since we are going to update GCC now, I think C++14 will only add clutter to the languages dropdown.

db/languages.yml
puqeko: Good spot! My legendary spelling and grammar skills at it again.

puqeko:
> Is there a reason to include C++14 rather than skip directly to C++17?
> I only created #42 (Add C++14) as an intermediate step until we could update GCC and add C++17. Since we are going to update GCC now, I think C++14 will only add clutter to the languages dropdown.

@tom93 , we weren't sure. It's up for debate. There is a good case for why not as it's less things to keep track of and support. However, some features are removed between C++14 and C++17 and C++14 is what they have been using at the IOI's as far as I'm aware https://ioi2019.az/en-content-26.html#COMMANDS.

Changes from 14 to 17: https://isocpp.org/files/papers/p0636r0.html
Changes from 11  to 14: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1319r0.html

db/languages.yml
puqeko: Thanks Chris :)

tom93:
> > Is there a reason to include C++14 rather than skip directly to C++17?
> > I only created #42 (Add C++14) as an intermediate step until we could update GCC and add C++17. Since we are going to update GCC now, I think C++14 will only add clutter to the languages dropdown.
> 
> @tom93 , we weren't sure. It's up for debate. There is a good case for why not as it's less things to keep track of and support. However, some features are removed between C++14 and C++17 and C++14 is what they have been using at the IOI's as far as I'm aware https://ioi2019.az/en-content-26.html#COMMANDS.
> 
> Changes from 14 to 17: https://isocpp.org/files/papers/p0636r0.html
> Changes from 11 to 14: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1319r0.html
> 
> The changes might be small enough that we don't care. What do you think?

Hmm, I didn't notice that features were removed. I suppose we can add C++14 for now, and remove it from the dropdown after the IOI switches to C++17.

puqeko:
Has been deployed. However,

- I added a commit to make `c11` the current version of `c`.
- Editing a submission seems to be broken.

tom93:
Are you referring to the fact that edits to the language of a submission are ignored? It also happens before this PR, so it's not related. (I might try to debug later.)

puqeko:
Yes, I think I’ve found the bug. I’ll start another PR.

PR #46, opened 2020-01-18
Update submit page

puqeko:
Improve readability and add details about the compilers used. Note that this change assumes PR #45 is merged.

2020-01-18 at 10 43 24 PM

### DIFF
--- a/app/views/problems/submit.html.erb
+++ b/app/views/problems/submit.html.erb
@@ -8,8 +8,9 @@
     </ul>
   </div>
 <% end %>
+<h2>Upload file</h2>
 <%= form_for @submission, :html => {:multipart => true}, :url => submit_problem_path(@problem) do |f| %>
-  <div class="field">
+  <div class="subfield">
     Language: 
     <%= f.select :language_id, Language.submission_options %>
   </div>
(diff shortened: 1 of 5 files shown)


app/views/problems/submit.html.erb
```diff
@@ -8,33 +8,48 @@
...
+<br>
+<hr>
+<br>
+<h2> Language details </h2>
+<p>
+<ul>
+  <li>C++ is compiled with g++ from GCC 9.2.1</li>
+  <li>C is compiled with gcc from GCC 9.2.1</li>
+  <li>Java uses javac 1.8 and OpenJDK 9</li>
+  <li>Haskell is compiled with GHC 7.10.3</li>
```
tom93: I don't think we should claim such precise versions when they can change in a routine 'apt upgrade'.
How about we only specify the major versions? GCC 9 and GHC 7.10 should be detailed enough for most people.
Another idea is to modify the compilation commands so they also print the compiler's version...

app/views/problems/submit.html.erb
```diff
@@ -8,33 +8,48 @@
...
-
+<br>
+<hr>
+<br>
+<h2> Language details </h2>
+<p>
+<ul>
+  <li>C++ is compiled with g++ from GCC 9.2.1</li>
+  <li>C is compiled with gcc from GCC 9.2.1</li>
+  <li>Java uses javac 1.8 and OpenJDK 9</li>
```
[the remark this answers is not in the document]
puqeko: Ah whoops :)

app/views/problems/submit.html.erb
puqeko: Agreed. I like the compilation command idea too. I'll leave this PR open for now to see if anyone wants to give implementing that a go

tom93: Maybe split off the readability improvements into a separate PR so they don't get held up by the language details?
I think they are pretty important, since the textarea input ("paste") is now visible to all users and in my experience the previous UI can be confusing.

puqeko: Done

PR #47, opened 2020-01-18
Update java

puqeko:
Update openjdk to version 11, the most recent stable release with LTS.

Server commands:
```
sudo add-apt-repository ppa:openjdk-r/ppa
sudo apt-get update
sudo apt install  openjdk-11-jdk
```

### DIFF
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -54,16 +54,22 @@ java:
   :lexer: java
   :compiled: yes
   :interpreted: yes
-  :current: java6
+  :current: java11
   :extension: .java
   :source_filename: Main
   :exe_extension: .jar
   :processes: 0
   :variants:
     java6:
-      :name: Java 1.6
+      :name: Java 1.6  # deprecated
       :compiler: /usr/bin/javac;/usr/bin/jar
       :compiler_command: '%{compiler[0]} -version | head -n 1 1>&2 && %{compiler[0]} -O -source 1.6 -J-Xms16m -J-Xmx256m %{source} && %{compiler[1]} cfe %{output} Main -C %{source_dir} Main.class'
+      :interpreter: /usr/bin/java
+      :interpreter_command: '%{interpreter} -jar %{source}'
+    java11:
+      :name: Java 11
+      :compiler: /usr/bin/javac;/usr/bin/jar
+      :compiler_command: '%{compiler[0]} -version | head -n 1 1>&2 && %{compiler[0]} -O -source 11 -J-Xms16m -J-Xmx256m %{source} && %{compiler[1]} cfe %{output} Main -C %{source_dir} Main.class'
       :interpreter: /usr/bin/java
       :interpreter_command: '%{interpreter} -jar %{source}'
 haskell:
@@ -128,4 +134,3 @@ j:
     j8:
       :name: "J 8.0.3"
       :interpreter: /usr/bin/ijconsole-8.0.3
-
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -69,7 +69,7 @@ java:
     java11:
       :name: Java 11
       :compiler: /usr/bin/javac;/usr/bin/jar
-      :compiler_command: '%{compiler[0]} -version | head -n 1 1>&2 && %{compiler[0]} -O -source 11 -J-Xms16m -J-Xmx256m %{source} && %{compiler[1]} cfe %{output} Main -C %{source_dir} Main.class'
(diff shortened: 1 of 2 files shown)


db/languages.yml
```diff
@@ -43,18 +43,24 @@ java:
...
-  :current: java6
+  :current: java11
   :extension: .java
   :source_filename: Main
   :exe_extension: .jar
   :processes: 0
   :variants:
     java6:
-      :name: Java 1.6
+      :name: Java 1.6  # depreciated
```
tom93:
```suggestion
      :name: Java 1.6  # deprecated
```

tom93:
Rebased and incorporated #50.

puqeko:
Has been deployed and seems to work

db/languages.yml
```diff
@@ -69,7 +69,7 @@ java:
     java11:
       :name: Java 11
       :compiler: /usr/bin/javac;/usr/bin/jar
-      :compiler_command: '%{compiler[0]} -version | head -n 1 1>&2 && %{compiler[0]} -O -source 11 -J-Xms16m -J-Xmx256m %{source} && %{compiler[1]} cfe %{output} Main -C %{source_dir} Main.class'
+      :compiler_command: '%{compiler[0]} -version | head -n 1 1>&2 && %{compiler[0]} -O -J-Xms16m -J-Xmx256m %{source} && %{compiler[1]} cfe %{output} Main -C %{source_dir} Main.class'
```
tom93: Why remove the source flag? If we update the JDK we'll probably want this "java11" variant to stay at the same version (and add a separate variant for the latest version).

puqeko: It didn't work when I deployed it. `-source 11` is an invalid option apparently. I am confused about the whole versioning thing. The source is supposed to be the language syntax version, which makes sense to fix at a number. If someone else knows more about this I'd be grateful if they could look into it

tom93: It works fine for me:
```
chroot$ javac -version
javac 11.0.5
chroot$ javac -source 11 Main.java
(succeeds)
```
---
Some notes about compatibility and '-source':

Java has pretty strong backward compatibility, so old working submissions should continue to work with newer JDK versions. (The only relevant incompatibility that I am aware of is a [corner case in String.split()](https://stackoverflow.com/questions/22718744/why-in-java-8-split-sometimes-removes-empty-strings-at-start-of-result-array) that changed in Java 8.)

I think we would also like broken submissions to remain broken if possible (e.g. if a submission's language is set to Java 6 but it tries to use Java 11 features, it should get a compilation error even after we install JDK 11).

This is where the '-source' option comes in. It indeed controls the language syntax version (e.g. '-source 1.6' does not allow Java 7's ["diamond" operator](https://docs.oracle.com/javase/tutorial/java/generics/types.html#diamond) and Java 8's lambda expressions).

However, it does not control the standard library version. So for example submissions that use the method [String.isBlank](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank()) (introduced in Java 11) will compile even when '-source 1.6' is used.

So '-source' is a very crude way to emulate older versions of Java without installing older JDKs.

---

While experimenting, I noticed that '-source 1.6' will be removed in the future:
```
chroot$ javac -version
javac 11.0.5
chroot$ javac -source 1.6 Main.java
warning: [options] bootstrap class path not set in conjunction with -source 6
warning: [options] source value 6 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
3 warnings
```
I'll add a note about this to db/languages.yml:
```
:name: Java 1.6  # deprecated (note: uses '-source 1.6', which will be removed in a future JDK release)
```
When we update to a JDK that doesn't support '-source 1.6' we can just change the option to '-source 1.7' (or some later version). Old Java 6 submissions that had compilation errors because they tried to use Java 7 syntax might start compiling successfully, but it's not a big deal. (There are some cases where code compiles under JDK 6 but not JDK 7, e.g. the sections "Improved Exception Handling" and "Method Type-Inference and Overload Resolution" in the [JDK 7 Compatibility Guide](https://www.oracle.com/technetwork/java/javase/compatibility-417013.html#incompatibilities), but they are very obscure.)

puqeko: Hmmm. Tried with the full compile command and all seems fine as you say. When I tested it after deploying it definitely gave an error durning compile. I wonder if either I didn’t update the dB properly or had an error in my submission. Either way, it seems so misdiagnosed the problem and it’s safe to add ‘-source 11’ back in.

PR #49, opened 2020-01-18
Add JavaScript

puqeko:
Compared Nodejs and V8 engine. Nodejs displayed much slower I/O than Python and the syntax for receiving input is horrible. V8 provides `readline` and much more competitive performance. Hence, we propose to support JavaScript by compiling the V8 engine from source and installing it in the sandbox. Has been tested on a local install of nztrain.

JavaScript is a popular language likely to remain relevant for the foreseeable future. It is commonly taught in schools to a basic level. I have received a few requests for us to support it.

**Note: The updated install script has not been tested.**

This has already been done on codeforces. https://codeforces.com/blog/entry/10024
https://v8.dev/docs/build
https://remusao.github.io/posts/compiling-v8-ubuntu-16-04.html

### DIFF
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -138,3 +138,17 @@ j:
     j8:
       :name: "J 8.0.3"
       :interpreter: /usr/bin/ijconsole-8.0.3
+javascript:
+  :name: JavaScript
+  :lexer: javascript
+  :compiled: no
+  :interpreted: yes
+  :current: v8_8
+  :extension: .js
+  :source_filename: program
+  :interpreter_command: '%{interpreter} %{source}'
+  :variants:
+    v8_8:
+      :name: "JavaScript (V8 8.1)"
+      :interpreter: /usr/bin/d8
+      :processes: 9
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -150,5 +150,5 @@ javascript:
   :variants:
     v8_8:
       :name: "JavaScript (V8 8.1)"
-      :interpreter: /usr/bin/d8
+      :interpreter: /usr/local/bin/d8
       :processes: 9
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -171,7 +171,7 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-[ -z "$TRAVIS" ] && { # if not in Travis-CI
+[ -z "$TRAVIS" ] && bash script/confirm.bash 'Install JavaScript V8 (submissions in V8 will fail without this!)' && {
   # JavaScript V8 engine compile and install
   echo "Preparing to compile and install the JavaScript V8 engine."
   echo
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -171,17 +171,8 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-[ -z "$TRAVIS" ] && bash script/confirm.bash 'Install V8 JavaScript Engine (submissions in JavaScript will fail without this!)' && {
-  echo "Preparing to compile and install the V8 JavaScript Engine."
-  echo
-  echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
-  echo "This may take a while..."
-  echo
-
+[ -z "$TRAVIS" ] && bash script/confirm.bash 'Install the V8 JavaScript Engine (submissions in JavaScript will fail without this!)' && {
   bash script/install/v8.bash
-
-  echo "JavaScript V8 installed into $ISOLATE_ROOT/usr/local/bin/d8"
-  echo
 }
 
 echo 'Finished chroot installs!'
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -170,7 +170,7 @@ chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g
 # gcc 9 done
 
 [ -z "$TRAVIS" ] && bash script/confirm.bash 'Install the V8 JavaScript Engine (submissions in JavaScript will fail without this!)' && {
-  chroot "$ISOLATE_ROOT" bash < script/install/v8.bash
+  HOME=/root chroot "$ISOLATE_ROOT" bash < script/install/v8.bash
 }
 
 umount "$ISOLATE_ROOT/proc"
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -177,39 +177,9 @@ chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g
   echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
   echo "This may take a while..."
   echo
-  set -x
 
-  : Making temporary directory
-  mkdir ~/v8-build
-  cd ~/v8-build
+  bash script/install/v8.bash
 
-  : Installing depot_tools
-  git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
-  export PATH="$PATH:$PWD/depot_tools"
-  gclient
-
-  : Getting V8 source code
-  mkdir v8
-  cd v8
-  fetch v8
-  cd v8
-
-  : Installing additional build dependencies
-  gclient sync
-  ./build/install-build-deps.sh
-
-  : Compiling V8
-  ./tools/dev/v8gen.py x64.release -- v8_use_external_startup_data=false
-  ninja -C out.gn/x64.release d8
-
-  : Copying executable into place
-  cp ./out.gn/x64.release/d8 "$ISOLATE_ROOT"/usr/bin/d8
-
-  : Cleaning up
-  cd ~
-  rm -r v8-build
-
-  set +x
   echo "JavaScript V8 installed into $ISOLATE_ROOT/usr/bin/d8"
   echo
 }
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -171,79 +171,48 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-# JavaScript V8 engine compile and install
-echo "Preparing to compile and install the JavaScript V8 engine."
-echo ""
-echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
-echo "This may take a while ..."
-echo ""
+[ -z "$TRAVIS" ] && { # if not in Travis-CI
+  # JavaScript V8 engine compile and install
+  echo "Preparing to compile and install the JavaScript V8 engine."
+  echo
+  echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
+  echo "This may take a while ..."
+  echo
+  set -x
 
-echo "Making temporary directory"
-echo "$chroot_cmd cd /"
-chroot "$ISOLATE_ROOT" cd /
-echo "$chroot_cmd mkdir v8-tools"
-chroot "$ISOLATE_ROOT" mkdir v8-tools
-echo "$chroot_cmd cd v8-tools"
-chroot "$ISOLATE_ROOT" cd v8-tools
-echo ""
+  : Making temporary directory
+  mkdir ~/v8-build
+  cd ~/v8-build
 
-echo "Getting V8 source"
-echo "$chroot_cmd git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git"
-chroot "$ISOLATE_ROOT" git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
-echo "$chroot_cmd export PATH=`pwd`/depot_tools:\"$PATH\""
-chroot "$ISOLATE_ROOT" export PATH=`pwd`/depot_tools:"$PATH"
-echo "$chroot_cmd cd .."
-chroot "$ISOLATE_ROOT" cd ..
-echo "$chroot_cmd mkdir v8"
-chroot "$ISOLATE_ROOT" mkdir v8
-echo "$chroot_cmd cd v8"
-chroot "$ISOLATE_ROOT" cd v8
-echo "$chroot_cmd fetch v8"
-chroot "$ISOLATE_ROOT" fetch v8
-echo ""
+  : Installing depot_tools
+  git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+  export PATH="$PATH:$PWD/depot_tools"
+  gclient
 
-echo "Syncing and checking out"
-echo "$chroot_cmd cd v8"
-chroot "$ISOLATE_ROOT" cd v8
-echo "$chroot_cmd gclient sync"
-chroot "$ISOLATE_ROOT" gclient sync
-echo ""
+  : Getting V8 source code
+  mkdir v8
+  cd v8
+  fetch v8
+  cd v8
 
-echo "Installing dependencies needed to build V8"
-echo "$chroot_cmd ./build/install-build-deps.sh"
-chroot "$ISOLATE_ROOT" ./build/install-build-deps.sh
+  : Installing additional build dependencies
+  gclient sync
+  ./build/install-build-deps.sh
 
-echo "Syncing again"
-echo "$chroot_cmd git pull origin master"
-chroot "$ISOLATE_ROOT" git pull origin master
-echo "$chroot_cmd gclient sync"
-chroot "$ISOLATE_ROOT" gclient sync
-echo ""
+  : Compiling V8
+  ./tools/dev/v8gen.py x64.release -- v8_use_external_startup_data=false
+  ninja -C out.gn/x64.release d8
 
-echo "Compile time!"
-echo "$chroot_cmd ./tools/dev/gm.py x64.release"
-chroot "$ISOLATE_ROOT" ./tools/dev/gm.py x64.release
-echo ""
+  : Copying executable into place
+  cp ./out.gn/x64.release/d8 "$ISOLATE_ROOT"/usr/bin/d8
 
-echo "Moving executable into place"
-echo "$chroot_cmd mv ./out/x64.release/d8 /usr/bin/d8"
-chroot "$ISOLATE_ROOT" mv ./out/x64.release/d8 /usr/bin/d8
-echo "$chroot_cmd mv ./out/x64.release/snapshot_blob.bin /usr/bin/snapshot_blob.bin"
-chroot "$ISOLATE_ROOT" mv ./out/x64.release/snapshot_blob.bin /usr/bin/snapshot_blob.bin
-echo ""
+  : Cleaning up
+  cd ~
+  rm -r v8-build
 
-echo "Removing source files"
-echo "$chroot_cmd cd /"
-chroot "$ISOLATE_ROOT" cd /
-echo "$chroot_cmd rm -r v8"
-chroot "$ISOLATE_ROOT" rm -r v8
-echo "$chroot_cmd rm -r v8-tools"
-chroot "$ISOLATE_ROOT" rm -r v8-tools
+  set +x
+  echo "JavaScript V8 installed into $ISOLATE_ROOT/usr/bin/d8"
+  echo
+}
 
-echo "Yay! JavaScript hopefully installed."
-echo "Do /usr/bin/d8 from chroot to test it out."
-echo ""
-# JavaScript done
-
-# let user know that chroot installs are finished
-echo "Finished chroot installs!"
+echo 'Finished chroot installs!'
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -170,7 +170,7 @@ chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g
 # gcc 9 done
 
 [ -z "$TRAVIS" ] && bash script/confirm.bash 'Install the V8 JavaScript Engine (submissions in JavaScript will fail without this!)' && {
-  bash script/install/v8.bash
+  chroot "$ISOLATE_ROOT" bash < script/install/v8.bash
 }
 
 umount "$ISOLATE_ROOT/proc"
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -171,12 +171,11 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-[ -z "$TRAVIS" ] && bash script/confirm.bash 'Install JavaScript V8 (submissions in V8 will fail without this!)' && {
-  # JavaScript V8 engine compile and install
-  echo "Preparing to compile and install the JavaScript V8 engine."
+[ -z "$TRAVIS" ] && bash script/confirm.bash 'Install V8 JavaScript Engine (submissions in JavaScript will fail without this!)' && {
+  echo "Preparing to compile and install the V8 JavaScript Engine."
   echo
   echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
-  echo "This may take a while ..."
+  echo "This may take a while..."
   echo
   set -x
 
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -171,5 +171,79 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-# let user know that chroot installs are finished
+# JavaScript V8 engine compile and install
+echo "Preparing to compile and install the JavaScript V8 engine."
+echo ""
+echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
+echo "This may take a while ..."
+echo ""
 
+echo "Making temporary directory"
+echo "$chroot_cmd cd /"
+chroot "$ISOLATE_ROOT" cd /
+echo "$chroot_cmd mkdir v8-tools"
+chroot "$ISOLATE_ROOT" mkdir v8-tools
+echo "$chroot_cmd cd v8-tools"
+chroot "$ISOLATE_ROOT" cd v8-tools
+echo ""
+
+echo "Getting V8 source"
+echo "$chroot_cmd git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git"
+chroot "$ISOLATE_ROOT" git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+echo "$chroot_cmd export PATH=`pwd`/depot_tools:\"$PATH\""
+chroot "$ISOLATE_ROOT" export PATH=`pwd`/depot_tools:"$PATH"
+echo "$chroot_cmd cd .."
+chroot "$ISOLATE_ROOT" cd ..
+echo "$chroot_cmd mkdir v8"
+chroot "$ISOLATE_ROOT" mkdir v8
+echo "$chroot_cmd cd v8"
+chroot "$ISOLATE_ROOT" cd v8
+echo "$chroot_cmd fetch v8"
+chroot "$ISOLATE_ROOT" fetch v8
+echo ""
+
+echo "Syncing and checking out"
+echo "$chroot_cmd cd v8"
+chroot "$ISOLATE_ROOT" cd v8
+echo "$chroot_cmd gclient sync"
+chroot "$ISOLATE_ROOT" gclient sync
+echo ""
+
+echo "Installing dependencies needed to build V8"
+echo "$chroot_cmd ./build/install-build-deps.sh"
+chroot "$ISOLATE_ROOT" ./build/install-build-deps.sh
+
+echo "Syncing again"
+echo "$chroot_cmd git pull origin master"
+chroot "$ISOLATE_ROOT" git pull origin master
+echo "$chroot_cmd gclient sync"
+chroot "$ISOLATE_ROOT" gclient sync
+echo ""
+
+echo "Compile time!"
+echo "$chroot_cmd ./tools/dev/gm.py x64.release"
+chroot "$ISOLATE_ROOT" ./tools/dev/gm.py x64.release
+echo ""
+
+echo "Moving executable into place"
+echo "$chroot_cmd mv ./out/x64.release/d8 /usr/bin/d8"
+chroot "$ISOLATE_ROOT" mv ./out/x64.release/d8 /usr/bin/d8
+echo "$chroot_cmd mv ./out/x64.release/snapshot_blob.bin /usr/bin/snapshot_blob.bin"
+chroot "$ISOLATE_ROOT" mv ./out/x64.release/snapshot_blob.bin /usr/bin/snapshot_blob.bin
+echo ""
+
+echo "Removing source files"
+echo "$chroot_cmd cd /"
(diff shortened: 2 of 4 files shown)


puqeko:
@Holmes98, could you please add the `language.yaml` stuff for this one?

Holmes98:
I think we should also add support for Node.js. While V8 is more performant, I doubt that many students/schools will want to build it from source (official binaries aren't available). Also, online compilers such as repl.it only support Node.js so it would be difficult for students to test their solutions locally.

puqeko:
> I think we should also add support for Node.js. While V8 is more performant, I doubt that many students/schools will want to build it from source (official binaries aren't available). Also, online compilers such as repl.it only support Node.js so it would be difficult for students to test their solutions locally.

Valid points. However, I'm still going to advocate that V8 is the lesser evil.

- The callback style philosophy for I/O in node is not great - particularly for our use case (example below). Maybe that's just me though.
- I presume that most students learn traditional javascript in the context of front-end web development. In which case, the ES6 syntax and interpreter will be somewhat alien to them anyway.
- We have to maintain another interpreter.
- Having NZIC contests solvable in all languages is a nice goal to have. It means everyone starting out in the NZIC are (as much as possible) on equal footings. It is less likely students will get confused and start to doubt the their chosen language instead of their own algorithms (I have received a few emails were this has been that case). However, adding Node.js to the mix will make this goal much more unrealistic/challenging to achieve. Bested on some quick tests the I/O is considerably slower than Python.

However, I agree that it is unrealistic for students to run their solutions locally with V8. To that end, it might not be worth having V8 unless we can recommend/provide a way for them to develop/test solutions.

Possible ways forward:

- We add Node.js anyway and leave it up to the students to work it out.
- We can develop some sort of way to run test code in our sandbox (lots of work maybe). But at least schools/students don't need to have Node.js or V8. Might be useful if we are to consider C# support as well.

Neither of these options seem overally enticing at the moment.

```
const readline = require('readline');
const rl = readline.createInterface({
    input: process.stdin,
    terminal: false
});

rl.on('line', (line) => {
    process.stdout.write(line + "\n");
});
```

puqeko:
The reason I think we should care so much about speed and usability is because I have a feeling JavaScript could become a very popular choice for NZIC participants. Currently we only really care about C++ and Python

Holmes98:
> Having NZIC contests solvable in all languages is a nice goal to have.

While this would indeed be nice, it seems unrealistic even with V8 – as far as I can tell JavaScript does not have built-in support for a few essential data structures (queues and priority queues) which would be very likely to come up at some point in NZIC.

I think the difference in I/O performance is unlikely to matter as algorithmic complexity should be the main factor in determining whether a solution is fast enough. I ran further tests on programs which do pretty much anything other than I/O (in this case summing 2 million integers) and Python was considerably slower than both variants of JavaScript.

```
$ time ./a.out < input  # C++
999724000624

real	0m0.525s
user	0m0.512s
sys	0m0.013s
$ time nodejs node.js < input
999726000624

real	0m0.536s
user	0m0.486s
sys	0m0.056s
$ time d8 v8.js < input
999724000624

real	0m0.411s
user	0m0.401s
sys	0m0.013s
$ time python3 test.py < input
999724000624

real	0m2.123s
user	0m1.961s
sys	0m0.162s
```

```python
# test.py
n = int(input())
total = 0
for i in range(n):
  total += int(input())
print(total)
```
```javascript
// v8.js
var n = parseInt(readline());
var total = 0;
for(var i = 0; i < n; i++) {
    total += parseInt(readline());
}
print(total);
```
```javascript
// node.js
const readline = require('readline');
const rl = readline.createInterface({input: process.stdin});
var total = 0;

rl.on('line', (line) => {
  total += parseInt(line, 10);
});

rl.on('close', () => {
  console.log(total);
});
```

puqeko:
It seems that output is the sticking point. Your results contradicted my initial tinkering so I compared your "summing integers" test to my "printing integers" test. Results as follows for N = 5000000:

### Summing ints
**Python**  real: 0m10.464s, user: 0m9.012s, sys: 0m1.448s
**Node** real: 0m1.469s, user: 0m1.352s, sys: 0m0.152s
**V8** real: 0m1.728s, user: 0m1.632s, sys: 0m0.104s

### Printing ints
**Python** real: 0m5.220s, user: 0m5.060s, sys: 0m0.156s
**Node** real: 0m28.581s, user: 0m19.700s, sys: 0m9.104s
**V8** real: 0m10.760s, user: 0m3.764s, sys: 0m6.992s

The time node takes can be halved by using `process.stdout.write` instead of `console.log`.

Holmes98:
So I've discovered (thanks to @lambda-fairy) that [Try It Online](https://tio.run/#javascript-v8) supports V8/d8 interpretation – we may not need to add Node.js after all.

db/languages.yml
```diff
@@ -138,3 +138,17 @@ j:
...
+  :lexer: javascript
+  :compiled: no
+  :interpreted: yes
+  :current: v8_8
+  :extension: .js
+  :source_filename: program
+  :interpreter_command: '%{interpreter} %{source}'
+  :variants:
+    v8_8:
+      :name: "JavaScript V8"
```
puqeko: Could be good to be more explicit with `V8 Engine 8.1`. Otherwise it looks like JavaScript version 8.

script/install/debootstrap.bash
```diff
@@ -171,5 +171,48 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+echo
+echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
+echo "This may take a while ..."
+echo
+set -x
+
+: Making temporary directory
+mkdir ~/v8-tools
+cd ~/v8-tools
+echo
```
puqeko: Is this echo intentional?

Holmes98: Oops, I missed deleting that one.

db/languages.yml
Holmes98: Will change this to `JavaScript (V8 8.1)` for now.

script/install/debootstrap.bash
```diff
@@ -171,5 +171,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+: Cleaning up
+cd ~
+rm -r v8-tools
+
+set +x
+echo "Yay! JavaScript hopefully installed."
+echo "Do /usr/bin/d8 from chroot to test it out."
+echo
+
+echo "Finished chroot installs!"
```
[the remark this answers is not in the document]
tom93: And I think we can be a bit more optimistic in the output message :)
e.g.
```
echo "JavaScript installed into $ISOLATE_ROOT/usr/bin/d8"
```
(by the time we deploy this it should have been tested enough; we shouldn't have to ask the person running the install script to test it out themselves)

script/install/debootstrap.bash
```diff
@@ -171,5 +171,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-# let user know that chroot installs are finished
-
+# JavaScript V8 engine compile and install
+echo "Preparing to compile and install the JavaScript V8 engine."
+echo
+echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
+echo "This may take a while ..."
```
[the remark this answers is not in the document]
Holmes98: Good point, it actually hits the job time limit on Travis.
https://travis-ci.org/NZOI/nztrain/jobs/640430838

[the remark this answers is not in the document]
tom93: Since it takes so long I think it's pretty important that we also have a way to disable V8 in normal installations. I'm not sure what's the best way to do this (we would also need to remove JavaScript from the language dropdown).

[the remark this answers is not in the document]
lambda-fairy: Travis set a [`$CI` environment variable](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) which we can branch on. I think that can take us half-way.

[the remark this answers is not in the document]
Holmes98: I've tested it on an 8-core VM and it the V8 build took about 15 minutes, does that still seem too long?

> we would also need to remove JavaScript from the language dropdown

If we're only disabling V8 on development installs I think it's fine to leave it (maybe with a message that "submissions in JavaScript V8 will not work").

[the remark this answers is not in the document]
Holmes98: I tried re-installing [from this ](https://github.com/NZOI/nztrain/commits/add-javascript-into-master) on a fresh Ubuntu 16.04 install and the whole process took 46 minutes.
- openjdk-11 took 20 minutes to download from ppa.launchpad.net (at ~230 kB/s).
- V8 took 7 minutes to download and another 7 minutes to build (using 8 cores).

If we want to add options to not install those I think that could be part of a separate pull request? In either case it should be pretty easy to add if we use [`confirm.bash`](https://github.com/NZOI/nztrain/blob/master/script/confirm.bash).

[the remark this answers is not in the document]
puqeko: I agree this should be part of a separate pull request. A longer term solution could be to only install GCC (or another language) by default. A separate install script can optionally be involved for all other languages - which means any future languages we add will also be left out of travis by default. The extra install step would need to be added to the README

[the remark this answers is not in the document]
Holmes98: For now, I can confirm that the following change works.
```diff
diff --git a/script/install/debootstrap.bash b/script/install/debootstrap.bash
index c07dba5..6fe92b9 100644
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -103,8 +103,10 @@ if ! chroot "$ISOLATE_ROOT" apt-cache show openjdk-11-jdk &>/dev/null; then
   chroot "$ISOLATE_ROOT" apt-get update
 fi
 
-echo "$chroot_install openjdk-11-jdk"
-chroot "$ISOLATE_ROOT" apt-get install openjdk-11-jdk # Java
+bash script/confirm.bash "Install Java (submissions in Java will fail without this!)" && {
+  echo "$chroot_install openjdk-11-jdk"
+  chroot "$ISOLATE_ROOT" apt-get install openjdk-11-jdk # Java
+}
 
 [ -z "$TRAVIS" ] && { # if not in Travis-CI
 
@@ -179,7 +181,7 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-[ -z "$TRAVIS" ] && { # if not in Travis-CI
+[ -z "$TRAVIS" ] && bash script/confirm.bash "Install JavaScript V8 (submissions in V8 will fail without this!)" && {
   # JavaScript V8 engine compile and install
   echo "Preparing to compile and install the JavaScript V8 engine."
   echo
```

Since the changes are small I don't mind adding it here.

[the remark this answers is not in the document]
tom93: > I've tested it on an 8-core VM and the V8 build took about 15 minutes, does that still seem too long?

For me it takes one hour on a single-core VM. Using `confirm.bash` will do, I don't mind if it's in this pull request or a separate one.

> > we would also need to remove JavaScript from the language dropdown
>
> If we're only disabling V8 on development installs I think it's fine to leave it (maybe with a message that "submissions in JavaScript V8 will not work").

Fair enough. If people want to deploy without JavaScript they can manually edit app/model/language.rb.

> openjdk-11 took 20 minutes to download from ppa.launchpad.net (at ~230 kB/s).

We should try to improve that. I'll create a separate pull request.

script/install/debootstrap.bash
```diff
@@ -171,5 +171,48 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  echo "Preparing to compile and install the JavaScript V8 engine."
+  echo
+  echo "If this fails, visit https://v8.dev/docs/build and follow the instructions."
+  echo "This may take a while ..."
+  echo
+  set -x
+
+  : Making temporary directory
+  mkdir ~/v8-build
+  cd ~/v8-build
```
tom93: Is there a reason to build V8 outside chroot? I think it would be cleaner to to keep everything related to the judging environment inside chroot (specifically, `./build/install-build-deps.sh` installs packages system-wide).

Running a sequence of commands inside chroot can be painful (e.g. `cd` doesn't persist across commands), but there are some approaches to make it easier:
 - Run bash in chroot and redirect the commands from a file (`chroot "$ISOLATE_ROOT" bash < install-v8.sh`) or from a [here-document](https://en.wikipedia.org/wiki/Here_document#Unix_shells)
 - [Export a function](https://stackoverflow.com/a/30792189)

script/install/debootstrap.bash
```diff
@@ -171,5 +171,48 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+
+  : Installing depot_tools
+  git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+  export PATH="$PATH:$PWD/depot_tools"
+  gclient
```
tom93:
```suggestion
  gclient --version # finishes the installation
```
This has the same effect, but doesn't print the long help message.
We could even remove this line completely; it's only required on Windows according to the [depot_tools tutorial](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/depot_tools_tutorial.html#_setting_up), and on Linux the `fetch` command will also do the required work. However, I'd like to keep the line to match the instructions (at least loosely).

script/install/debootstrap.bash
puqeko: We had issues with `git` I believe

script/install/debootstrap.bash
```diff
@@ -171,5 +171,48 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  : Compiling V8
+  ./tools/dev/v8gen.py x64.release -- v8_use_external_startup_data=false
+  ninja -C out.gn/x64.release d8
+
+  : Copying executable into place
+  cp ./out.gn/x64.release/d8 "$ISOLATE_ROOT"/usr/bin/d8
+
+  : Cleaning up
+  cd ~
+  rm -r v8-build
```
tom93:
```suggestion
  rm -rf v8-build
```
For me it needed `-f` to remove read-only files. Did it work for you without `-f`?

Holmes98: It did work without `-f` for me but I wouldn't mind adding it.

script/install/debootstrap.bash
Holmes98: I was also unsure if it's necessary but the instructions [here](https://v8.dev/docs/source-code) seemed to imply that it should be done on both Windows and Linux.

script/install/debootstrap.bash
tom93: From private discussion it turns to be my fault, I was running the commands directly instead of using `install.bash` (which runs `debootstrap.bash` as root, and then `rm` doesn't prompt about read-only files). I still suggest `-f` in case we want to run these commands as non-root.

script/install/debootstrap.bash
Holmes98: I'm currently trying to get V8 building inside chroot using the first method (see 49feb87) but am getting errors. Any ideas?

```
+ ./build/install-build-deps.sh
sudo: unable to resolve host holmes-nztrain2
sudo: unable to resolve host holmes-nztrain2
Hit:1 http://nz.archive.ubuntu.com/ubuntu xenial InRelease
Hit:2 http://nz.archive.ubuntu.com/ubuntu xenial-updates InRelease                                         
Get:3 http://nz.archive.ubuntu.com/ubuntu xenial/multiverse i386 Packages [140 kB]                                                   
Get:4 http://nz.archive.ubuntu.com/ubuntu xenial-updates/multiverse i386 Packages [15.9 kB]                                          
Hit:5 http://archive.ubuntu.com/ubuntu xenial InRelease                                                                              
Get:6 http://security.ubuntu.com/ubuntu xenial-security InRelease [109 kB]                    
Hit:7 http://ppa.launchpad.net/brightbox/ruby-ng/ubuntu xenial InRelease
Get:8 http://archive.ubuntu.com/ubuntu xenial-updates InRelease [109 kB]                                                             
Hit:9 http://ppa.launchpad.net/deadsnakes/ppa/ubuntu xenial InRelease                                              
Hit:10 http://ppa.launchpad.net/openjdk-r/ppa/ubuntu xenial InRelease                                                                
Get:11 http://archive.ubuntu.com/ubuntu xenial/main i386 Packages [1196 kB]                                                
Get:12 http://security.ubuntu.com/ubuntu xenial-security/main i386 Packages [631 kB]                      
Hit:13 http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu xenial InRelease                                    
Get:14 http://ppa.launchpad.net/brightbox/ruby-ng/ubuntu xenial/main i386 Packages [8532 B]                                          
Get:15 http://ppa.launchpad.net/deadsnakes/ppa/ubuntu xenial/main i386 Packages [31.3 kB]                                            
Get:16 http://ppa.launchpad.net/openjdk-r/ppa/ubuntu xenial/main i386 Packages [11.8 kB]                                             
Get:17 http://security.ubuntu.com/ubuntu xenial-security/restricted i386 Packages [7224 B]                                           
Get:18 http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu xenial/main i386 Packages [85.7 kB]
Get:19 http://archive.ubuntu.com/ubuntu xenial/universe i386 Packages [7512 kB]                                                      
Get:20 http://archive.ubuntu.com/ubuntu xenial-updates/main i386 Packages [892 kB]                                                   
Get:21 http://archive.ubuntu.com/ubuntu xenial-updates/restricted i386 Packages [7580 B]                                             
Fetched 10.8 MB in 41s (259 kB/s)                                                                                                    
Reading package lists... Done
Building apt package list.
Skipping debugging symbols.
Including 32-bit libraries.
Including ARM cross toolchain.
Including NaCl, NaCl toolchain, NaCl ports dependencies.
Finding missing packages...
Packages required:  libasound2:i386 libcap2:i386 libelf-dev:i386 libfontconfig1:i386 libglib2.0-0:i386 libgpm2:i386 libncurses5:i386 libnss3:i386 libpango1.0-0:i386 libpci3:i386 libssl-dev:i386 libssl1.0.0:i386 libtinfo-dev:i386 libudev1:i386 libuuid1:i386 libx11-xcb1:i386 libxcomposite1:i386 libxcursor1:i386 libxdamage1:i386 libxi6:i386 libxrandr2:i386 libxss1:i386 libxtst6:i386 linux-libc-dev:i386 ant apache2-bin autoconf binutils binutils-aarch64-linux-gnu binutils-arm-linux-gnueabihf binutils-mips64el-linux-gnuabi64 binutils-mipsel-linux-gnu bison bzip2 cdbs cmake curl dbus-x11 devscripts dpkg-dev elfutils fakeroot flex g++-5-multilib g++-5-multilib-arm-linux-gnueabihf g++-arm-linux-gnueabihf g++-mingw-w64-i686 gawk gcc-5-multilib-arm-linux-gnueabihf gcc-arm-linux-gnueabihf git-core gperf intltool lib32gcc1 lib32ncurses5-dev lib32stdc++6 lib32z1-dev libapache2-mod-php7.0 libappindicator3-1 libappindicator3-dev libasound2 libasound2-dev libatk1.0-0 libatspi2.0-0 libatspi2.0-dev libav-tools libbluetooth-dev libbrlapi-dev libbrlapi0.6 libbz2-1.0 libbz2-dev libc6 libc6-dev libc6-dev-armhf-cross libc6-i386 libcairo2 libcairo2-dev libcap-dev libcap2 libcups2 libcups2-dev libcurl4-gnutls-dev libdrm-dev libdrm2 libelf-dev libevdev-dev libevdev2 libexpat1 libffi-dev libffi6 libfontconfig1 libfreetype6 libgbm-dev libgbm1 libglib2.0-0 libglib2.0-dev libglu1-mesa-dev libgnome-keyring-dev libgnome-keyring0 libgtk-3-0 libgtk-3-dev libinput-dev libinput10 libjpeg-dev libkrb5-dev libnspr4 libnspr4-dev libnss3 libnss3-dev libpam0g libpam0g-dev libpango1.0-0 libpci-dev libpci3 libpcre3 libpixman-1-0 libpng16-16 libpulse-dev libpulse0 libsctp-dev libspeechd-dev libspeechd2 libsqlite3-0 libsqlite3-dev libssl-dev libstdc++6 libtinfo-dev libtool libudev-dev libudev1 libuuid1 libvulkan-dev libvulkan1 libwayland-egl1-mesa libwww-perl libx11-6 libx11-xcb1 libxau6 libxcb1 libxcomposite1 libxcursor1 libxdamage1 libxdmcp6 libxext6 libxfixes3 libxi6 libxinerama1 libxkbcommon-dev libxrandr2 libxrender1 libxslt1-dev libxss-dev libxt-dev libxtst-dev libxtst6 linux-libc-dev-armhf-cross locales openbox p7zip patch perl php7.0-cgi pkg-config python python-cherrypy3 python-crypto python-dev python-numpy python-opencv python-openssl python-psutil python-yaml realpath rpm ruby subversion texinfo uuid-dev wdiff x11-utils xcompmgr xsltproc xutils-dev xvfb xz-utils zip zlib1g

E: Unable to correct problems, you have held broken packages.
The following command failed:  apt-get --just-print install libasound2:i386 libcap2:i386 libelf-dev:i386 libfontconfig1:i386 libglib2.0-0:i386 libgpm2:i386 libncurses5:i386 libnss3:i386 libpango1.0-0:i386 libpci3:i386 libssl-dev:i386 libssl1.0.0:i386 libtinfo-dev:i386 libudev1:i386 libuuid1:i386 libx11-xcb1:i386 libxcomposite1:i386 libxcursor1:i386 libxdamage1:i386 libxi6:i386 libxrandr2:i386 libxss1:i386 libxtst6:i386 linux-libc-dev:i386 ant apache2-bin autoconf binutils binutils-aarch64-linux-gnu binutils-arm-linux-gnueabihf binutils-mips64el-linux-gnuabi64 binutils-mipsel-linux-gnu bison bzip2 cdbs cmake curl dbus-x11 devscripts dpkg-dev elfutils fakeroot flex g++-5-multilib g++-5-multilib-arm-linux-gnueabihf g++-arm-linux-gnueabihf g++-mingw-w64-i686 gawk gcc-5-multilib-arm-linux-gnueabihf gcc-arm-linux-gnueabihf git-core gperf intltool lib32gcc1 lib32ncurses5-dev lib32stdc++6 lib32z1-dev libapache2-mod-php7.0 libappindicator3-1 libappindicator3-dev libasound2 libasound2-dev libatk1.0-0 libatspi2.0-0 libatspi2.0-dev libav-tools libbluetooth-dev libbrlapi-dev libbrlapi0.6 libbz2-1.0 libbz2-dev libc6 libc6-dev libc6-dev-armhf-cross libc6-i386 libcairo2 libcairo2-dev libcap-dev libcap2 libcups2 libcups2-dev libcurl4-gnutls-dev libdrm-dev libdrm2 libelf-dev libevdev-dev libevdev2 libexpat1 libffi-dev libffi6 libfontconfig1 libfreetype6 libgbm-dev libgbm1 libglib2.0-0 libglib2.0-dev libglu1-mesa-dev libgnome-keyring-dev libgnome-keyring0 libgtk-3-0 libgtk-3-dev libinput-dev libinput10 libjpeg-dev libkrb5-dev libnspr4 libnspr4-dev libnss3 libnss3-dev libpam0g libpam0g-dev libpango1.0-0 libpci-dev libpci3 libpcre3 libpixman-1-0 libpng16-16 libpulse-dev libpulse0 libsctp-dev libspeechd-dev libspeechd2 libsqlite3-0 libsqlite3-dev libssl-dev libstdc++6 libtinfo-dev libtool libudev-dev libudev1 libuuid1 libvulkan-dev libvulkan1 libwayland-egl1-mesa libwww-perl libx11-6 libx11-xcb1 libxau6 libxcb1 libxcomposite1 libxcursor1 libxdamage1 libxdmcp6 libxext6 libxfixes3 libxi6 libxinerama1 libxkbcommon-dev libxrandr2 libxrender1 libxslt1-dev libxss-dev libxt-dev libxtst-dev libxtst6 linux-libc-dev-armhf-cross locales openbox p7zip patch perl php7.0-cgi pkg-config python python-cherrypy3 python-crypto python-dev python-numpy python-opencv python-openssl python-psutil python-yaml realpath rpm ruby subversion texinfo uuid-dev wdiff x11-utils xcompmgr xsltproc xutils-dev xvfb xz-utils zip zlib1g

It produced the following output:
Reading package lists...
Building dependency tree...
Reading state information...
fakeroot is already the newest version (1.20.2-1ubuntu1).
fakeroot set to manually installed.
libcap2 is already the newest version (1:2.24-12).
libffi-dev is already the newest version (3.2.1-4).
libffi-dev set to manually installed.
libffi6 is already the newest version (3.2.1-4).
libpcre3 is already the newest version (2:8.38-3.1).
libtinfo-dev is already the newest version (6.0+20160213-1ubuntu1).
libtinfo-dev set to manually installed.
libxau6 is already the newest version (1:1.0.8-1).
libxau6 set to manually installed.
libxcb1 is already the newest version (1.11.1-1ubuntu1).
libxcb1 set to manually installed.
libxdmcp6 is already the newest version (1:1.1.2-1.1).
libxdmcp6 set to manually installed.
libxext6 is already the newest version (2:1.3.3-1).
libxext6 set to manually installed.
xz-utils is already the newest version (5.1.1alpha+20120614-2ubuntu2).
xz-utils set to manually installed.
binutils is already the newest version (2.26.1-1ubuntu1~16.04.8).
binutils set to manually installed.
dpkg-dev is already the newest version (1.18.4ubuntu1.6).
dpkg-dev set to manually installed.
libc6 is already the newest version (2.23-0ubuntu11).
libc6-dev is already the newest version (2.23-0ubuntu11).
libc6-dev set to manually installed.
libglib2.0-0 is already the newest version (2.48.2-0ubuntu4.4).
libglib2.0-0 set to manually installed.
libx11-6 is already the newest version (2:1.6.3-1ubuntu2.1).
libx11-6 set to manually installed.
patch is already the newest version (2.7.5-1ubuntu0.16.04.2).
patch set to manually installed.
perl is already the newest version (5.22.1-9ubuntu0.6).
perl set to manually installed.
python is already the newest version (2.7.12-1~16.04).
python set to manually installed.
ruby is already the newest version (2:2.3.0+1bbox1~xenial1).
libstdc++6 is already the newest version (9.2.1-17ubuntu1~16.04).
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 binutils-mips64el-linux-gnuabi64 : Depends: binutils (= 2.26-8ubuntu2) but 2.26.1-1ubuntu1~16.04.8 is to be installed
 binutils-mipsel-linux-gnu : Depends: binutils (= 2.26-8ubuntu2) but 2.26.1-1ubuntu1~16.04.8 is to be installed
 elfutils : Depends: libdw1 (= 0.165-3ubuntu1) but 0.165-3ubuntu1.2 is to be installed
            Depends: libelf1 (= 0.165-3ubuntu1) but 0.165-3ubuntu1.2 is to be installed
 g++-5-multilib-arm-linux-gnueabihf : Depends: gcc-5-arm-linux-gnueabihf-base (= 5.3.1-14ubuntu2cross1) but 5.4.0-6ubuntu1~16.04.9cross1 is to be installed
                                      Depends: g++-5-arm-linux-gnueabihf (= 5.3.1-14ubuntu2cross1)
                                      Depends: libsfstdc++-5-dev-armhf-cross (= 5.3.1-14ubuntu2cross1) but it is not going to be installed or
                                               libsfstdc++-5-dev-armhf-cross but it is not going to be installed
 g++-arm-linux-gnueabihf : Depends: g++-5-arm-linux-gnueabihf (>= 5.3.1-3~)
 gcc-5-multilib-arm-linux-gnueabihf : Depends: gcc-5-arm-linux-gnueabihf-base (= 5.3.1-14ubuntu2cross1) but 5.4.0-6ubuntu1~16.04.9cross1 is to be installed
                                      Depends: gcc-5-arm-linux-gnueabihf (= 5.3.1-14ubuntu2cross1)
 gcc-arm-linux-gnueabihf : Depends: gcc-5-arm-linux-gnueabihf (>= 5.3.1-3~)

You will have to install the above packages yourself.
```

puqeko: Just guess, but a quick google of `sudo: unable to resolve host' gave this https://askubuntu.com/questions/59458/error-message-sudo-unable-to-resolve-host-none.

Maybe check in `/etc/hosts` that `holmes-nztrain` has an entry under localhost. On the train site there is no entry in `/etc/hosts` for `train.nzoi.org.nz`.

tom93: I get the same error about `binutils-mips64el-linux-gnuabi64` dependencies, still investigating.

Holmes98: Adding the hostname to `/etc/hosts` gets rid of the `sudo: unable to resolve host` message but the dependency errors still persist.

tom93: It looks like debootstrap.bash creates an inconsistent /etc/apt/sources.list:
https://github.com/NZOI/nztrain/blob/42a9f42fad99acb1343cf1c71b7754e754474861/script/install/debootstrap.bash#L34-L40

Adding "universe" to `http://security.ubuntu.com/ubuntu $SUITE-security` fixes the problem.
I'll create a pull request to clean it up.

Holmes98: I've tried building V8 inside chroot again with the above fix but ran into a few more issues:
- Python 2 needs to be installed otherwise `gclient` fails ([Python 2 is currently not being installed](https://github.com/NZOI/nztrain/pull/62))
```
+ gclient --version
/v8-build/depot_tools/gclient: line 32: exec: python: not found
```
- `fetch v8` fails. It seems that `gclient` isn't creating the required files. At some point I was able to get this working but I can't seem to reproduce the fix.
```
+ fetch v8
/v8-build/depot_tools/vpython: line 42: /v8-build/depot_tools/.cipd_bin/vpython: No such file or directory
```
- During the one time I managed to get the above working there was still a single unmet dependency when running `./build/install-build-deps.sh`, though the build still appeared to complete successfully so I'm unsure if it's actually needed.
```
The following packages have unmet dependencies:
 python-numpy : Depends: liblapack3 but it is not going to be installed or
                         liblapack.so.3

You will have to install the above packages yourself.
```

Holmes98: So the issue with `fetch v8` seems to be fixable by moving `umount "$ISOLATE_ROOT/proc"` to after the V8 build under `debootstrap.bash`. Also note that `wget` needs to be installed else it gives the same error (`wget` is already installed under `debootstrap.bash` but I had it commented out while testing to make things faster).

Holmes98: See 2488faa for a working build inside chroot. I didn't get the missing dependency this time but it's quite a bit slower inside chroot since there are more uninstalled build dependencies to download (takes me about 50 minutes to download+build). Also there are a bunch of `No such file or directory` errors/warnings ([see the build log here](https://pastebin.com/raw/1SpBzgBd)).

I'm not sure whether it'd be better to build V8 inside or outside chroot.

tom93: Can you please try the following flags to reduce download size:
```diff
-git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+git clone --depth 1 https://chromium.googlesource.com/chromium/tools/depot_tools.git

-fetch v8
+fetch --no-history v8

-gclient sync
+gclient sync --no-history

-./build/install-build-deps.sh
+./build/install-build-deps.sh --no-syms --no-arm --no-chromeos-fonts --no-nacl
```
(for me the last one reduced apt downloads in chroot from 400 MB to 244 MB).
If it's still too much we can build it outside chroot, but let's write the script in a way that makes it easier run in chroot later.

Regarding the locale warnings, `export LC_ALL=C` fixes them.

script/install/debootstrap.bash
```diff
@@ -171,5 +171,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+: Installing additional build dependencies
+gclient sync
+./build/install-build-deps.sh
+
+: Compiling V8
+./tools/dev/v8gen.py x64.release -- v8_use_external_startup_data=false
+ninja -C out.gn/x64.release d8
+
+: Moving executable into place
+mv ./out.gn/x64.release/d8 "$ISOLATE_ROOT"/usr/bin/d8
```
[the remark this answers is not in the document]
tom93: Actually, I take it back -- I don't think copying `d8` instead of moving it is going to help much. On the other hand, something that I think _will_ help is if the script exited on errors (`set -e`) instead of continuing. Otherwise if something goes wrong (e.g. build error, or failure to copy `d8`) the script will still delete the entire directory, so everything would have to be downloaded from scratch.

script/install/debootstrap.bash
Holmes98: I can confirm that this works (reduces the build to 26 minutes for me). I don't mind whether we build it inside or outside chroot.

Holmes98:
Looks fine to me. I don't mind building in chroot, so let's just go with that then.

tom93:
Deployed!

PR #50, opened 2020-01-18
Show compiler version info

puqeko:
Assumes #46. See comments

### DIFF
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -7,22 +7,28 @@ c++:
   :lexer: c++
   :compiled: yes
   :interpreted: no
-  :current: c++11
+  :current: c++17
   :extension: .cpp
   :source_filename: program
   :exe_extension: .exe
   :processes: 1
   :variants:
-    c++03:
-      :name: "C++03"
+    c++03:  # depreciated
+      :name: "C++03 (GCC 9)"
       :compiler: /usr/bin/g++
       :compiler_command: "%{compiler} -std=gnu++03 -O2 -o %{output} %{source} -lm"
-#      upgrade: c++11 # submissions will be upgraded if possible, if support is removed, it will be force-upgraded
-#      note: probably will never happen for this particular example... (or at least not for the next 10 years)
     c++11:
-      :name: "C++11"
+      :name: "C++11 (GCC 9)"
       :compiler: /usr/bin/g++
       :compiler_command: "%{compiler} -std=gnu++11 -O2 -o %{output} %{source} -lm"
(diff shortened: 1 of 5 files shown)


db/languages.yml
```diff
@@ -7,24 +7,30 @@ c++:
...
   :processes: 1
   :variants:
-    c++03:
+    c++03:  # depreciated
       :name: "C++03"
       :compiler: /usr/bin/g++
-      :compiler_command: "%{compiler} -std=gnu++03 -O2 -o %{output} %{source} -lm"
-#      upgrade: c++11 # submissions will be upgraded if possible, if support is removed, it will be force-upgraded
-#      note: probably will never happen for this particular example... (or at least not for the next 10 years)
+      :compiler_command: "%{compiler} --version 1>&2 && %{compiler} -std=gnu++03 -O2 -o %{output} %{source} -lm"
```
tom93: How about we only show the first line of the version (without the copyright)?
```
"%{compiler} --version | head -n 1 >&2 && ..."
```

db/languages.yml
```diff
@@ -52,7 +63,7 @@ java:
     java6:
       :name: Java 1.6
       :compiler: /usr/bin/javac;/usr/bin/jar
-      :compiler_command: '%{compiler[0]} -O -source 1.6 -J-Xms16m -J-Xmx256m %{source} && %{compiler[1]} cfe %{output} Main -C %{source_dir} Main.class'
+      :compiler_command: '%{compiler[0]} -version 1>&2 && %{compiler[1]} --version 1>&2 && %{compiler[0]} -O -source 1.6 -J-Xms16m -J-Xmx256m %{source} && %{compiler[1]} cfe %{output} Main -C %{source_dir} Main.class'
```
tom93: I don't think anybody cares about the version of `jar`.

db/languages.yml
puqeko: Good idea

db/languages.yml
puqeko: Fair enough

tom93:
The submission page shows the compilation command. I think ideally it should exclude the part that displays the version, but it's not important and can be changed later.

puqeko:
> The submission page shows the compilation command. I think ideally it should exclude the part that displays the version, but it's not important and can be changed later.

Prehaps there needs to be another field in the .yml for a version command which can be run on startup. For now, this should do

puqeko:
2020-01-19 at 12 24 50 PM

puqeko:
Has been deployed successfully

PR #52, opened 2020-01-19
Sort language dropdown and set default to C++17

Holmes98:
Currently the default option is the first in the dropdown, which is C11.

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -51,11 +51,20 @@ class Language < ActiveRecord::Base
     {key => path}.merge Hash[(0...elements.size).map{|i|"#{key}[#{i}]".to_sym}.zip(elements)]
   end
 
+  # for validating submissions
   def self.submission_options
-    latest = LanguageGroup.where(identifier: %w[c++ c python haskell java ruby j]).pluck(:current_language_id)
-    old = Language.where(identifier: %w[c++11 c++14 c99]).pluck(:id)
-    languages = Language.where(:id => latest).order(:identifier) + Language.where(:id => old).order(:identifier)
-    Hash[languages.map{ |language| ["#{language.name}", language.id] }]
+    grouped_submission_options.values.reduce(:merge)
+  end
+
+  # for the selection dropdown
+  def self.grouped_submission_options
+    current = ["c++", "c", "python", "java", "haskell", "ruby", "j"]  # language group identifiers, order is preserved
+    other = ["c++14", "c++11", "c99"]  # language identifiers, ordered by language name
+    current_langs = LanguageGroup.where(:identifier => current).preload(:current_language).map(&:current_language)
+    current_options = current_langs.sort_by { |lang| current.index(lang.group.identifier) }.map { |lang| [lang.name, lang.id] }
+    other_options = Language.where(:identifier => other).order(:name).pluck(:name, :id)
+    { "Current" => Hash[current_options],
+      "Other" => Hash[other_options] }
   end
 
   def self.default
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -58,6 +58,10 @@ class Language < ActiveRecord::Base
     Hash[languages.map{ |language| ["#{language.name}", language.id] }]
   end
 
+  def self.default
+    LanguageGroup.find_by(identifier: 'c++').current_language
+  end
+
   def self.infer(ext)
     case ext
     when *%w[.cpp]
--- a/app/views/problems/submit.html.erb
+++ b/app/views/problems/submit.html.erb
@@ -12,7 +12,7 @@
 <%= form_for @submission, :html => {:multipart => true}, :url => submit_problem_path(@problem) do |f| %>
   <div class="subfield">
     Language: 
-    <%= f.select :language_id, Language.submission_options %>
+    <%= f.select :language_id, options_for_select(Language.submission_options, @submission.language_id || Language.default.id) %>
   </div>
   <div class="field">
     Source file: 
@@ -30,7 +30,7 @@
   <%= form_for @submission, :url => submit_problem_path(@problem) do |f| %>
     <div class="subfield">
       Language: 
-      <%= f.select :language_id, Language.submission_options %>
+      <%= f.select :language_id, options_for_select(Language.submission_options, @submission.language_id || Language.default.id) %>
     </div>
     <div class="subactions">
       <%= f.submit 'Submit' %>
--- a/app/views/problems/submit.html.erb
+++ b/app/views/problems/submit.html.erb
@@ -12,7 +12,7 @@
 <%= form_for @submission, :html => {:multipart => true}, :url => submit_problem_path(@problem) do |f| %>
   <div class="subfield">
     Language: 
-    <%= f.select :language_id, options_for_select(Language.submission_options, @submission.language_id || Language.default.id) %>
+    <%= f.select :language_id, grouped_options_for_select(Language.grouped_submission_options, @submission.language_id || Language.default.id) %>
   </div>
   <div class="field">
     Source file: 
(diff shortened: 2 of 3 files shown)


app/models/language.rb
```diff
@@ -54,10 +54,14 @@ def explode_path_to_hash key, path
   def self.submission_options
     latest = LanguageGroup.where(identifier: %w[c++ c python haskell java ruby j]).pluck(:current_language_id)
     old = Language.where(identifier: %w[c++11 c++14 c99 python2]).pluck(:id)
-    languages = Language.where(:id => latest).order(:identifier) + Language.where(:id => old).order(:identifier)
+    languages = Language.where(:id => latest + old).order(:name)
```
tom93: I think it was good having the old languages at the bottom (though a separator might help).
Sorting by name rather than identifier makes sense since that is what the user sees, but I think the order will still be the same because the identifiers and names are similar, right?

Holmes98: I think it looks cleaner when sorted fully, and it makes it a bit easier to find a particular option – here's a comparison (I also removed C99): 

image image

Sorting by name because when/if we add JavaScript V8 I'm thinking of having the name starting with `JavaScript` but the variant name as something like `v8_8`.

tom93: Most people shouldn't use the old languages. I agree that it's currently hard to find a particular option, but I think a separator would mostly solve that problem without cluttering up the current languages. Python 3.4 is popular and moves from 6th down to 9th in the comparison (and if we wanted to keep C99 it would be even further down).

puqeko: Having the languages in order benefits those who have never looked at the language list before and are trying to find their preferred language - in which case, an alphabetical ordering is best. 

Everyone else is only interested in selecting their preferred language as quickly as possible - in which case, a popularity/relevance ordering is best.

It depends which case we are optimising for.

puqeko: My ideal ordering would be

C++17
Python 3.4
Java 1.6
Haskell 2010
Ruby 2.2
J 8.0.3
-----
Everything else

lambda-fairy: Have you considered visually separating the new and old languages?

Either using `optgroup` tags or [dividers](https://stackoverflow.com/a/4317135/617159).

puqeko: Not sure how that would translate into ruby. I can make it do this... But the divider is selectable (which isn't ideal). There must be a better way

2020-01-19 at 9 55 20 PM

lambda-fairy: I think you'll have to change the template that generates the HTML as well?

puqeko: 2020-01-19 at 10 45 05 PM

Ignoring the ordering, this is what using option groups looks like

lambda-fairy: Modulo that extra colon, that looks quite nice 🙂

What does everyone else think?

Holmes98: I like it as well. @puqeko should be adding this change to #54 shortly.

puqeko:
2020-01-19 at 10 45 05 PM

app/models/language.rb
```diff
@@ -51,11 +51,26 @@ def explode_path_to_hash key, path
...
-    old = Language.where(identifier: %w[c++11 c++14 c99 python2]).pluck(:id)
-    languages = Language.where(:id => latest).order(:identifier) + Language.where(:id => old).order(:identifier)
-    Hash[languages.map{ |language| ["#{language.name}", language.id] }]
+    grouped_submission_options.values.reduce(:merge)
+  end
+
+  # for the selection dropdown
+  def self.grouped_submission_options
+    current = ["c++", "c", "python", "java", "haskell", "ruby", "j"]
+    other = ["c++14", "c++11", "c99", "python2"]
```
tom93: Maybe add comments to draw attention to the fact that one contains language group identifiers and the other contains language identifiers?

app/models/language.rb
```diff
@@ -51,11 +51,26 @@ def explode_path_to_hash key, path
...
+    current_ids = LanguageGroup.where(:identifier => current).pluck(:current_language_id)
+    current_options = Language.where(:id => current_ids).to_a.sort_by { |lang| current.index(lang.group.identifier) }.map { |lang| [lang.name, lang.id] }
```
tom93: How about `LanguageGroup.where(:identifier => current).map(&:current_language)`? It's less efficient because it loads the entire LanguageGroup model (rather than just the current_language_id column) and makes a separate query for each Language (could be fixed using `.preload(:current_language).map...`), but it's cleaner.

Another approach for preserving the order (not sure which I prefer):
```
current_langs_by_ident = LanguageGroup.where(:identifier => current).map(&:current_language).index_by { |lang| lang.group.identifier }
current_options = current.map { |ident| current_langs_by_ident[ident] }.map { |lang| [lang.name, lang.id] }
```

app/models/language.rb
```diff
@@ -51,11 +51,26 @@ def explode_path_to_hash key, path
...
+    other_options = Language.where(:identifier => other).to_a.sort_by { |lang| lang.name }.map { |lang| [lang.name, lang.id] }
```
tom93: I prefer `.order(:name)`, even though it's inconsistent with the previous line (could then also use `.pluck(:name, :id)`).

app/models/language.rb
tom93: I like the optgroups too.

app/models/language.rb
Holmes98: Like this, or is the line getting too long?
```
current_options = LanguageGroup.where(:identifier => current).preload(:current_language).map(&:current_language).to_a.sort_by { |lang| current.index(lang.group.identifier) }.map { |lang| [lang.name, lang.id] }
```

tom93: I'd say it's too long. I was thinking
```
current_langs = LanguageGroup.where(:identifier => current).map(&:current_language)
current_options = current_langs.sort_by { |lang| current.index(lang.group.identifier) }.map { |lang| [lang.name, lang.id] }
```
`.to_a` isn't necessary after `.map`, and I think it could also be left out in the original version, i.e. `.where(...).sort_by`

Holmes98:
Have rebased to add @tom93's suggestions and clean up the commit history.

app/models/language.rb
puqeko: What does the `&` in `&:current_language` do? It's probably obvious that I don't know ruby. My current strategy is pretty much "hit it until it works and hope to learn something along the way". Thanks for pointing these things out.

tom93: It [converts a proc to a block](https://ruby-doc.org/core-2.7.0/doc/syntax/calling_methods_rdoc.html#label-Proc+to+Block+Conversion).
`.map` expects a block argument, e.g. `.map { |g| g.current_language }`, so `&` is required to pass the symbol `current_language` as the block argument (rather than a normal positional argument).
Symbols are [convertible to procs](https://ruby-doc.org/core-2.7.0/Proc.html#class-Proc-label-Conversion+of+other+objects+to+procs), and "calling" them on an object calls the object's method with that name.

puqeko:
Could someone please rebase again from deployed so I can merge. I still can't figure out how to push to a forked repository :(

```
puqeko@mrtwiddles:~/Desktop/nztrain$ git push h h/sort-submissions-dropdown --force
Username for 'https://github.com': puqeko
Password for 'https://puqeko@github.com': 
Total 0 (delta 0), reused 0 (delta 0)
To https://github.com/Holmes98/nztrain.git
 ! [remote rejected] h/sort-submissions-dropdown -> h/sort-submissions-dropdown (permission denied)
error: failed to push some refs to 'https://github.com/Holmes98/nztrain.git'
```

puqeko:
2020-01-20 at 10 58 03 PM

tom93:
> commit 66e4393423cddee2576fa02bdca1ae0859acb2ec
> Requires `db:migrate`

This shouldn't require `db:migrate`. However, the earlier changes to `db/languages.yml` will require `db:seed` (which is safe to run on an existing database).

puqeko:
Ah, I've confused this PR with #54 again, which does need a db:migrate. I did one anyway but it probably didn't do anything. The seeds did happen earlier.

PR #53, opened 2020-01-19
Correctly record the new language selected for the updated submission

puqeko:
Currently, we updating (editing) a submission if you change the language then the submission will appear to retain the original language.


tom93:
Nice catch.

puqeko:
Deployed successfully

PR #54, opened 2020-01-19
Add default language per-user

Holmes98:
As an alternative option (or addition) to #52.

Adds a default language setting on registration and on the profile settings page. If a user hasn't set their default language, it uses `Language.default` which is C++17.


puqeko:
Covered by #52

puqeko:
Jokes

Holmes98:
Closing this as I have a better idea.

PR #55, opened 2020-01-19
Update haskell (using stack)

puqeko:
Because I know at least one student uses this and it is cool to have at-least one functional language.

```
curl -sSL https://get.haskellstack.org/ | sh
stack update
stack install ghc
apt-get remove ghc
apt-get autoremove
```

### DIFF
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -84,8 +84,13 @@ chroot "$ISOLATE_ROOT" apt-get install build-essential # C/C++ (g++, gcc)
 echo "$chroot_install ruby"
 chroot "$ISOLATE_ROOT" apt-get install ruby # Ruby (ruby)
 
-echo "$chroot_install ghc"
-chroot "$ISOLATE_ROOT" apt-get install ghc # Haskell (ghc)
+echo "Getting Stack for Haskell"
+echo "$chroot_cmd curl -sSL https://get.haskellstack.org/ | sh"
+chroot "$ISOLATE_ROOT" curl -sSL https://get.haskellstack.org/ | sh
+echo "$chroot_cmd stack update"
+chroot "$ISOLATE_ROOT" stack update
+echo "$chroot_cmd stack install ghc"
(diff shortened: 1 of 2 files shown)


script/install/debootstrap.bash
```diff
@@ -84,8 +84,13 @@ chroot "$ISOLATE_ROOT" apt-get install build-essential # C/C++ (g++, gcc)
...
 
-echo "$chroot_install ghc"
-chroot "$ISOLATE_ROOT" apt-get install ghc # Haskell (ghc)
+echo "Getting Stack for Haskell"
+echo "$chroot_cmd curl -sSL https://get.haskellstack.org/ | sh"
+chroot "$ISOLATE_ROOT" curl -sSL https://get.haskellstack.org/ | sh
+echo "$chroot_cmd stack update"
+chroot "$ISOLATE_ROOT" stack update
+echo "$chroot_cmd stack install ghc"
+chroot "$ISOLATE_ROOT" stack install ghc # Haskell (ghc)
```
tom93: These commansd install ghc into $HOME, which won't work. When a user runs the debootstrap script their $HOME will leak into the chroot environment, while isolate.rb sets $HOME to "/box".

puqeko:
So you reckon the best option is stick to this?

```
chroot "$ISOLATE_ROOT" add-apt-repository -y ppa:hvr/ghc
chroot "$ISOLATE_ROOT" apt-get install -y ghc-8.8.2
```

tom93:
I think so. Both have very recent versions of GHC; the install size with hvr is smaller (1GB vs 5GB); stack adds complexity because it installs into $HOME and is project-based. Are there reasons in favour of stack?
 
I created a separate pull request (#61) which uses hvr's PPA (not changing this one in case we want to try using stack in the future).

puqeko:
The reason for going with stack initially was because it would be easy to upgrade. However, I agree that installing it directly seems easier for version control and to reduce space. Wouldn't be hard to upgrade either. If we can get it working, may as well go the direct route

Holmes98:
Closing as we have merged #61 instead.

PR #27, opened 2019-02-02
Update GHC

leifmetcalf:
This uses hvr's very-well-maintained ppa (https://launchpad.net/~hvr/+archive/ubuntu/ghc). No idea how shell scripting works so something's probably wrong.

PR #61, opened 2020-01-21
Update Haskell (using hvr's PPA)

tom93:
This PPA is recommended by https://www.haskell.org/downloads/linux/.

An earlier attempt was made in #27.

(We should probably refactor the debootstrap script at some point to add all the PPAs at the start so it doesn't need to run `apt-get update` multiple times.)

 - [ ] Changes applied manually on server

### DIFF
--- a/app/services/isolate.rb
+++ b/app/services/isolate.rb
@@ -232,10 +232,10 @@ EOF
       'lib' => [], # core libraries
       'lib64' => [], # 64-bit libraries
       'usr' => [], # general binaries, includes and libraries
+      'opt' => [], # for ghc from ppa:hvr/ghc
       'etc' => [],
       #'etc/alternatives' => [], # required for many symbolic links to work
       #'etc/j' => [], # required for J to work (load profile)
-      'var/lib' => [] # /var/lib/ghc required by ghc (Haskell compiler)
     }.map do |dir, opt|
       fullpath = File.expand_path(dir, isolate_root)
       boxpath = File.expand_path(dir, "/")
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -86,7 +86,7 @@ haskell:
     haskell2010:
       :name: "Haskell 2010"
       :compiler: /usr/bin/ghc
-      :compiler_command: '%{compiler} --make -O2 -o %{output} %{source} -lm'
+      :compiler_command: '%{compiler} --version 1>&2 && %{compiler} --make -O2 -o %{output} %{source} -lm'
 python:
   :name: Python
   :lexer: python
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -83,8 +83,18 @@ chroot "$ISOLATE_ROOT" apt-get install build-essential # C/C++ (g++, gcc)
 echo "$chroot_install ruby"
 chroot "$ISOLATE_ROOT" apt-get install ruby # Ruby (ruby)
 
-echo "$chroot_install ghc"
-chroot "$ISOLATE_ROOT" apt-get install ghc # Haskell (ghc)
+# add haskell ppa
+echo "$chroot_cmd add-apt-repository ppa:hvr/ghc -y"
+chroot "$ISOLATE_ROOT" add-apt-repository ppa:hvr/ghc -y
+
+echo "$chroot_cmd apt-get update"
+chroot "$ISOLATE_ROOT" apt-get update
+
+echo "$chroot_install ghc-8.8.2"
+chroot "$ISOLATE_ROOT" apt-get install ghc-8.8.2 # Haskell (ghc)
+
+echo "$chroot_cmd update-alternatives --install /usr/bin/ghc ghc /opt/ghc/8.8.2/bin/ghc 75"
+chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/ghc ghc /opt/ghc/8.8.2/bin/ghc 75
 
 if ! chroot "$ISOLATE_ROOT" apt-cache show openjdk-11-jdk &>/dev/null; then
   # add java ppa
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -96,6 +96,13 @@ chroot "$ISOLATE_ROOT" apt-get install ruby # Ruby (ruby)
 
   echo "$chroot_cmd update-alternatives --install /usr/bin/ghc ghc /opt/ghc/8.8.2/bin/ghc 75"
   chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/ghc ghc /opt/ghc/8.8.2/bin/ghc 75
+
+  # Note: Running ghc requires /proc to be mounted. The isolate command mounts
+  # it, but it might not be mounted if running ghc manually in the chroot.
+  # (To find shared libraries, the ghc binary has a RUNPATH attribute with
+  # paths that are relative to $ORIGIN. The dynamic linker uses /proc/self/exe
+  # to expand $ORIGIN. It can be overridden using the environment variable
+  # LD_ORIGIN_PATH.)
 }
 
 if ! chroot "$ISOLATE_ROOT" apt-cache show openjdk-11-jdk &>/dev/null; then
--- a/script/install/debootstrap.bash
+++ b/script/install/debootstrap.bash
@@ -83,18 +83,20 @@ chroot "$ISOLATE_ROOT" apt-get install build-essential # C/C++ (g++, gcc)
 echo "$chroot_install ruby"
 chroot "$ISOLATE_ROOT" apt-get install ruby # Ruby (ruby)
 
-# add haskell ppa
-echo "$chroot_cmd add-apt-repository ppa:hvr/ghc -y"
-chroot "$ISOLATE_ROOT" add-apt-repository ppa:hvr/ghc -y
+[ -z "$TRAVIS" ] && { # if not in Travis-CI
+  # add haskell ppa
+  echo "$chroot_cmd add-apt-repository ppa:hvr/ghc -y"
+  chroot "$ISOLATE_ROOT" add-apt-repository ppa:hvr/ghc -y
 
-echo "$chroot_cmd apt-get update"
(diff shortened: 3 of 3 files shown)


puqeko:
Is it normal for `ghc` to install in `/opt/`? The required library is at  `./opt/ghc/8.8.2/lib/ghc-8.8.2/haskeline-0.7.5.0/libHShaskeline-0.7.5.0-ghc8.8.2.so` but `ghc` is searching in these paths:

```
     32484:	  trying file=/lib/x86_64-linux-gnu/tls/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/lib/x86_64-linux-gnu/tls/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/lib/x86_64-linux-gnu/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/lib/x86_64-linux-gnu/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/x86_64-linux-gnu/tls/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/x86_64-linux-gnu/tls/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/x86_64-linux-gnu/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/x86_64-linux-gnu/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/lib/tls/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/lib/tls/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/lib/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/lib/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/tls/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/tls/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/x86_64/libHShaskeline-0.7.5.0-ghc8.8.2.so
     32484:	  trying file=/usr/lib/libHShaskeline-0.7.5.0-ghc8.8.2.so
```

Prehaps we need to symlink it into `/usr/lib` or something?

puqeko:
Fixed by running

```
ln -fst /usr/lib/ /opt/ghc/8.8.2/lib/ghc-8.8.2/*/*.so
ln -fst /usr/lib/ /opt/ghc/8.8.2/lib/ghc-8.8.2/*/*.so.*
```

Surly there is a better way though.

tom93:
Nice, how did you get it to print the paths it was searching?
The [PPA docs](https://launchpad.net/~hvr/+archive/ubuntu/ghc) explicitly say they install to /opt/ghc. I think that's just a decision made by the package maintainer to allow multiple versions to be installed without conflicts. But I'm not sure how that relates to the dynamic linker configuration (and why it works outside the debootstrap chroot but not inside).
If this is not urgent, I'd like to keep trying to track down the root cause rather than resort to the /usr/lib symlink trick.

puqeko:
I used `LD_DEBUG=libs,files /opt/ghc/bin/ghc` to display debug information. It might be useful to compare this output to a version installed outside chroot.

lambda-fairy:
You can also use `ghc-pkg list` to show what packages GHC can see.

puqeko:
Ah neat. Thanks.

It seems it can't see any without symlinks...
```
root@train:/usr/lib# /opt/ghc/bin/ghc-pkg list
/opt/ghc/8.8.2/lib/ghc-8.8.2/bin/ghc-pkg: error while loading shared libraries: libHSterminfo-0.4.1.4-ghc8.8.2.so: cannot open shared object file: No such file or directory
```

But with symlinks...

```
root@train:/# /opt/ghc/bin/ghc-pkg list
/opt/ghc/8.8.2/lib/ghc-8.8.2/package.conf.d
    Cabal-3.0.1.0
    array-0.5.4.0
    base-4.13.0.0
    binary-0.8.7.0
    bytestring-0.10.10.0
    containers-0.6.2.1
    deepseq-1.4.4.0
    directory-1.3.4.0
    filepath-1.4.2.1
    ghc-8.8.2
    ghc-boot-8.8.2
    ghc-boot-th-8.8.2
    ghc-compact-0.1.0.0
    ghc-heap-8.8.2
    ghc-prim-0.5.3
    ghci-8.8.2
    haskeline-0.7.5.0
    hpc-0.6.0.3
    integer-gmp-1.0.2.0
    libiserv-8.8.2
    mtl-2.2.2
    parsec-3.1.14.0
    pretty-1.1.3.6
    process-1.6.7.0
    rts-1.0
    stm-2.5.0.0
    template-haskell-2.15.0.0
    terminfo-0.4.1.4
    text-1.2.4.0
    time-1.9.3
    transformers-0.5.6.2
    unix-2.7.2.2
    xhtml-3000.2.2.1
```

tom93:
I found the problem, the root cause was that /proc wasn't mounted.

The ghc binary has a RUNPATH attribute in the dynamic section to find the shared libraries (view using `objdump -p`), and uses $ORIGIN to specify paths relative to the binary (see [man ld.so](http://man7.org/linux/man-pages/man8/ld.so.8.html#DESCRIPTION)). The dynamic linker reads /proc/self/exe to determine the path to the binary, which fails when /proc isn't mounted (strace was helpful here).

I see a few ways to fix this:

1.  Mount the proc filesystem: `mount -t proc proc /proc`
    We would need to run this on every boot. I'm don't think there are security implications (on one hand it allows processes running in chroot to view e.g. the list of processes running outside chroot, but on the other hand submissions are executed by `isolate` which prevents that).

2.  Write a small wrapper for ghc that sets `LD_ORIGIN_PATH=/opt/ghc/8.8.2/lib/ghc-8.8.2/bin`
    This is my preferred option.

3.  Install the shared libraries system-wide

        printf '%s\n' /opt/ghc/8.8.2/lib/ghc-8.8.2/*/ > /etc/ld.so.conf.d/opt-ghc.conf
        ldconfig

puqeko:
Nice going. I prefer the second option too. However, if we put `LD_ORIGIN_PATH=/opt/ghc/8.8.2/lib/ghc-8.8.2/bin` in the compile command then it will be visible on the submission page. Or did you have another way in mind? I'm facing a similar dilemma with C# at the moment

tom93:
We can write a shell script that sets `LD_ORIGIN_PATH` then calls the real GHC, and use that as the compilation command. We can even install the shell script as /usr/bin/ghc to make it completely transparent.

puqeko:
I borrowed your idea for #64. It feels a bit dirty, but it will look clean on the user side and it wasn't too difficult to implement. Sounds like the best option for `ghc` as well.

puqeko:
Suggested changes to be consistent-ish with #64. You probably have a better way. Can make them into commits if you want.

`debootstrap.bash`
```
[ -z "$TRAVIS" ] && bash script/confirm.bash 'Install GHC 8.8.2 for Haskell' && {
  set +x
  if ! chroot "$ISOLATE_ROOT" apt-cache show ghc-8.8.2 &>/dev/null; then
    chroot "$ISOLATE_ROOT" add-apt-repository ppa:hvr/ghc -y
  fi
  chroot "$ISOLATE_ROOT" apt-get install ghc-8.8.2 # Haskell (ghc)

  : Moving ghc wrapper into place
  # see https://github.com/NZOI/nztrain/pull/61
  GHC_LD_ORIGIN_PATH=/opt/ghc/8.8.2/lib/ghc-8.8.2/bin \
  GHC_BINARY=/opt/ghc/8.8.2/bin/ghc \
  envsubst < script/haskell/compile-command-template.sh | cat > "$ISOLATE_ROOT/usr/local/ghc.sh"
  chroot "$ISOLATE_ROOT" ln "/usr/local/ghc.sh" "/usr/bin/ghc"
  chroot "$ISOLATE_ROOT" chmod +x "/usr/bin/ghc"

  set -x
  echo "GHC 8.8.2 for Haskell installed"
}
```

`script/haskell/compile-command-template.sh`
```
#!/bin/bash

LD_ORIGIN_PATH=$GHC_LD_ORIGIN_PATH $GHC_BINARY

exit ${PIPESTATUS[0]}
```

tom93:
It turns out that the isolate command mounts /proc, so there is no need for $LD_ORIGIN_PATH. I added a comment to explain.

I updated the isolate command to include /opt in the sandbox.

I decided not to use confirm.bash, because apt-get already prompts before installing.

For merging, I suggest squash-and-merge with the following message:

```
Update GHC to 8.8.2 using hvr's PPA (#61)

This PPA is recommended by https://www.haskell.org/downloads/linux/.

Since it installs ghc into /opt, add that to the isolate environment
(and remove /var/lib, which is no longer required by ghc).

Also update the compile command to display version info, and skip
installation on Travis-CI.
```

(or we can rebase onto 'deployed' using git and merge to keep the separate commits; but note that I have a typo in the message of commit 68fae7a, the version should be 8.8.2)

Holmes98:
Note that `apt-get` doesn't prompt for confirmation if there aren't any dependencies to install, which was the case for me.
```
ghc-8.8.2
  Depends: gcc
  Depends: libc6
  Depends: libc6-dev
  Depends: libgmp10
  Depends: libgmp-dev
  Depends: libncursesw5
  Depends: libtinfo-dev
```

Otherwise, I've tested this and everything looks good.

tom93:
Yes. Unfortunately there doesn't appear to be a way to force apt-get to prompt when there are no dependencies to install[1], but on clean systems I think it is very likely that some dependencies (specifically libgmp-dev) will need to be installed. Is it possible that in your case the dependencies were installed because you already had the old version of ghc?

[1] https://unix.stackexchange.com/questions/164464/force-apt-get-to-prompt-yes-no/522719

Holmes98:
True, I already had an older ghc installed. This should be fine then.

PR #63, opened 2020-01-23
Enable "universe" component of security repository in sources.list

tom93:
(in debootstrap script)

The "universe" component is enabled for the "$SUITE" repository, so it should also be enabled for "$SUITE-security" to get security updates.

Discovered because it caused dependency errors when building V8 (https://github.com/NZOI/nztrain/pull/49#discussion_r369930432).

The sources.list file could do with some more cleanup. Leaving it for now to reduce merge conflicts with commit 9f5ff36 ("Fix isolate root path in debootstrap script", 2019-01-30).

This will need to be applied manually on the server.


tom93:
Deployed and applied changes manually.

PR #56, opened 2020-01-19
Add C# (Mono)

puqeko:
Up for debate.

Should be relatively easy to add. We had some students at camp last year who were taught it in school. Not sure how popular it still is now that the standards have changed.

**Note: Hasn't been tested yet**

https://www.mono-project.com/download/stable/#download-lin

PR #64, opened 2020-01-23
Add C# (.Net Core)

puqeko:
Possible replacement for #56. However, it has been much more involved than Mono to get going.

Tested the install script on my machine and had C# compiling and running on train but a few issues came up.
- <strike>The `csc.dll` prints everything to stdout. This means that errors don't tell nztrain that compilation failed and instead qless crashes because there is no output file. Fixed by creating a shell script to redirect the version info to stdout and the errors to stderr.</strike> **_The shell script needed to return the exit status of the command._**
- <strike>The path to the buildtime reference assemblies has a more specific version number (3.1.101) than can be specified during install (3.1.1). Hence, it is possible this path will break if someone downloads a future 3.3.1. I couldn't figure out how to make the compiler command generic in bash.</strike> **_Fixed thanks to @Holmes98_**
- I'm not sure if I have included enough compile time references. It's just the one file - I tried executing a file which utilised `System.Collections` and it still worked so I presume all the fundamental stuff is available.
- The `dotnet` command needed to be given a `.json` file to tell which runtime framework version to use.

If someone could please check over that would be great.


puqeko:
Not a very descriptive commit message on my part (bac5179). I changed the shebang at some point and it turned out that isolate didn't like it so I've changed it back to `\bin\bash\`.

For some reason, the `.sh` file's output isn't being captured by isolate and I'm not sure why. The code in that area is a bit of a mess.

Consequences of not capturing output:

- If the code compiles and runs successfully it will work fine.
- If any errors occur then nztrain won't know and will continue until something breaks.
- Version info isn't displayed.

2020-01-24 at 11 48 45 AM

script/csharp/dotnet-config-template.json
```diff
@@ -0,0 +1,8 @@
+{
+    "runtimeOptions": {
+        "framework": {
+            "name": "$DOTNET_FW_NAME",
+            "version": "$DOTNET_FW_VERSION"
+        }
+    }
+}
```
Holmes98:
```suggestion
}

```

db/languages.yml
```diff
@@ -138,3 +138,21 @@ j:
...
+  :source_filename: program
+  :processes: 10
+  :interpreter: /usr/bin/csr  # custom shell script
+  :variants:
+    csharp8:
+      :name: "C# 8.0"
+      :compiler: /usr/bin/csc  # custom shell script
+      # Note: csc already prints the version number when executed
+      :compiler_command: "%{compiler} %{source} -optimize+ -out:%{output} -langversion:8.0"
+      :interpreter_command: '%{interpreter} %{source}'
```
Holmes98:
```suggestion
      :interpreter_command: '%{interpreter} %{source}'

```

script/install/debootstrap.bash
```diff
@@ -173,5 +173,49 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  UBUNTU_VERSION=$(lsb_release --release --short)
+  wget -q "https://packages.microsoft.com/config/ubuntu/${UBUNTU_VERSION}/packages-microsoft-prod.deb" -O packages-microsoft-prod.deb
+  dpkg -i packages-microsoft-prod.deb
+  rm packages-microsoft-prod.deb
+  ["$UBUNTU_VERSION" == "18.04"] && sudo add-apt-repository universe
+  apt-get install -y apt-transport-https
+  apt-get update
+
+  : Installing dotnet 3.1
+  apt-get install -y dotnet-sdk-3.1
```
Holmes98: These will need to be done inside chroot.

script/install/debootstrap.bash
```diff
@@ -173,5 +173,49 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+
+  : Creating runtime config file
+  # The dotnet command will not execute the compiled .exe unless there is a config file
+  # which tells it what runtime framework to use.
+  # https://stackoverflow.com/questions/46065777/is-it-possible-to-compile-a-single-c-sharp-code-file-with-the-net-core-roslyn-c
+
+  RUNTIME_CONFIG_PATH="/usr/share/dotnet-config.json"  # used in the compile command
```
Holmes98:
```suggestion
  RUNTIME_CONFIG_PATH="$ISOLATE_ROOT/usr/share/dotnet-config.json"  # used in the compile command
```

script/install/debootstrap.bash
```diff
@@ -173,5 +173,49 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+
+  : Installing .Net Core for C#
+  # https://docs.microsoft.com/en-us/dotnet/core/install/linux-package-manager-ubuntu-1604
+
+  : Downloading package info from Microsoft
+  UBUNTU_VERSION=$(lsb_release --release --short)
+  wget -q "https://packages.microsoft.com/config/ubuntu/${UBUNTU_VERSION}/packages-microsoft-prod.deb" -O packages-microsoft-prod.deb
+  dpkg -i packages-microsoft-prod.deb
+  rm packages-microsoft-prod.deb
+  ["$UBUNTU_VERSION" == "18.04"] && sudo add-apt-repository universe
```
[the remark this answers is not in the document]
puqeko: Interesting, thanks

script/install/debootstrap.bash
```diff
@@ -173,5 +173,49 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+
+  : Moving shell scripts into place
+  
+  cp "script/csharp/compile-command.sh" "/usr/local/csc.sh"
+  ln "/usr/local/csc.sh" "/usr/bin/csc"
+  chmod +x "/usr/bin/csc"
+
+  cp "script/csharp/run-command.sh" "/usr/local/csr.sh"
+  ln "/usr/local/csr.sh" "/usr/bin/csr"
+  chmod +x "/usr/bin/csr"
```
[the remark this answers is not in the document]
puqeko: Good catch. Shouldn't we also need `$ISOLATE_ROOT` in the first `ln` argument?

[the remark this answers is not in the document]
Holmes98: True

script/install/debootstrap.bash
puqeko: Whoops. Completely forgot about that. Will fix it up

db/languages.yml
```diff
@@ -138,3 +138,21 @@ j:
...
       :interpreter: /usr/bin/ijconsole-8.0.3
+
+csharp:
+  :name: "C#"
+  :lexer: csharp
+  :compiled: yes
+  :interpreted: yes
+  :current: csharp8
+  :extension: .cs
+  :source_filename: program
```
Holmes98:
```suggestion
  :source_filename: program
  :exe_extension: .exe
```
Needs this otherwise it gives a runtime error – `dotnet exec needs a managed .dll or .exe extension`.

script/csharp/compile-command.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# This script does two things
+#   - Redirects errors from stdout to stderr so that nztrain knows when compilation fails
+#   - Supply .dll references to the csc command
+
+# Print version and copyright by compiling nothing
+/usr/bin/dotnet /usr/share/dotnet/sdk/3.1.101/Roslyn/bincore/csc.dll | \
+# print the first line, which has compiler version info, to stdout
+head -n 1 >&1
```
Holmes98:
```suggestion
# print the first line, which has compiler version info, to stderr
head -n 1 1>&2
```
The submission page compilation output shows `stderr`, not `stdout`.

script/csharp/compile-command.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Actually compile:
+# [dotnet] [csc] -reference:[stdlib.dll] [...passed in args...]
+/usr/bin/dotnet /usr/share/dotnet/sdk/3.1.101/Roslyn/bincore/csc.dll \
+-reference:/usr/share/dotnet/sdk/3.1.101/ref/netstandard.dll "$@" | \
+# Remove the first 3 lines of output. The rest, if any, are errors.
+tail -n +4 >&2
```
Holmes98:
```suggestion
tail -n +4 1>&2

exit ${PIPESTATUS[0]}
```
- To be consistent with the other compiler commands in `languages.yml`.
- `exit` with the status code of the compilation so the judge knows if the compilation failed.

puqeko: Ah, this was the bit that I was missing. I erroneously figured that nztrain was using the stderr channel to determine if the compilation succeeded or failed. Great work, thanks!

Since this is the case, there is now no need to run the compiler command twice. I'll resolve these two requests with a single commit.

script/csharp/compile-command.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This script does two things
+#   1. Redirects all output to stderr
+#   2. Supplies .dll references to the csc command
+
+# Actually compile:
+# [dotnet] [csc] -reference:[stdlib.dll] [...passed in args...]
+/usr/bin/dotnet /usr/share/dotnet/sdk/3.1.101/Roslyn/bincore/csc.dll \
+-reference:/usr/share/dotnet/sdk/3.1.101/ref/netstandard.dll "$@" 1>&2
```
Holmes98:
```suggestion
version="$(dotnet --list-sdks | awk '{ print $1 }')"
/usr/bin/dotnet /usr/share/dotnet/sdk/"$version"/Roslyn/bincore/csc.dll \
-reference:/usr/share/dotnet/sdk/"$version"/ref/netstandard.dll "$@" 1>&2
```
Not sure if this is a good way to do it but it does work.

puqeko: Even better, we could just do `dotnet --version`?

Holmes98: Yes that does seem better... I'm not sure why I didn't try that.

script/csharp/compile-command.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
Holmes98:
```suggestion
#!/bin/bash
```
This should be bash if we're using `PIPESTATUS` (https://unix.stackexchange.com/questions/14270/get-exit-status-of-process-thats-piped-to-another).

puqeko: Okay, in that case both of them should be `bash`.

app/models/language.rb
```diff
@@ -58,7 +58,7 @@ def self.submission_options
 
   # for the selection dropdown
   def self.grouped_submission_options
-    current = ["c++", "c", "python", "java", "haskell", "ruby", "j"]  # language group identifiers, order is preserved
+    current = ["c++", "csharp", "c", "python", "java", "haskell", "ruby", "j"]  # language group identifiers, order is preserved
```
tom93:
```suggestion
    current = ["c++", "c", "python", "csharp", "java", "haskell", "ruby", "j"]  # language group identifiers, order is preserved
```
The "Current" languages are ordered by preference; I think C# should be further down (after Python).

Holmes98: In that case, should we also move Python above C?

tom93: I feel like C and C++ should be together...

puqeko: We should probably decide what to order on. If it is by "preference" then Python is more popular for NZIC followed by C++. Hardly anyone uses C so, although I agree that they usually belong together, it would make more sense under this scheme to have C further down.

puqeko: But it seems, at least, we all agree that Python should move ahead of C#. So that's something :)

tom93: Yes, I think everyone loves Python :)

I'm still not sure about moving C down. I was going to post the following as a separate comment, but it seems like it might be somewhat relevant to the current discussion:

> The language list is getting pretty long (bearing in mind we'll also add JavaScript), so searching will be difficult again (previous discussion: https://github.com/NZOI/nztrain/pull/52#discussion_r368271778). I'm starting to think that @Holmes98's original idea to sort the entire list might have been the right thing to do all along. There is no need to hold up this PR, but we should think about it. Some ideas:
>  - Move J to "Other" (or remove it completely; only Ronald has every used it)
>  - Move Ruby to "Other"
>  - Remove old versions: C99 (as @Holmes98 originally wanted to do), C++11 (we should keep C++14 for the IOI, but I can't think of a reason to keep C++11). This will help if we sort the entire list.

Holmes98: I'll later submit a PR that will set each user's default language to the language of their latest submission. For almost everyone this should remove the need to re-select their language on each submission, so I honestly don't really mind what order we have the languages in. The only issue is that a lot of people will then have their default as C++11 instead of C++17 (and might not notice the new version) so I was thinking of also removing C++11 at the same time.

puqeko: I second removing J.

Perhaps a better name for the "Other" group is "Older" or "Outdated"?

Since Python and C++ are the most popular by a mile, maybe we put Python and C++ first then sort the remaining languages in "Current" alphabetically. If we put Python first then C++ and C would conveniently appear next to each other.

puqeko: Otherwise, sorting each group would mean we don't have to keep having this discussion every time a new language is added. But I think there is value in separating out the outdated Ruby and C++14/11 versions while we have them

tom93: > I'll later submit a PR that will set each user's default language to the language of their latest submission.

Sounds great. The issue of not noticing when a newer version becomes available is annoying though, and will come up again every time we update the language.

Would it be too complicated/slow to add some logic along the lines of

"If the last submission's language is no longer the current version within its group, and the user has never submitted in the current version of the language, then use the current version of that language as the default."

This is meant to catch cases where a user tries C++17 but then intentionally wants to go back to C++14. But if a user is careful to never select C++17, then it will keep trying to set it as their default. The only solution I can think of is to also record the date when the language version was added, which seems too much hassle.

> Perhaps a better name for the "Other" group is "Older" or "Outdated"? In which case, Ruby fits (for now).

I thought that at first, but I started liking "Other" because it gave us the freedom to put less-preferred (but still up-to-date) languages there. So we could make C++ and Python into the "Current" languages, and everything else into "Other". (But then maybe rename "Current" to "Favourites" or something.)

puqeko: > I thought that at first, but I started liking "Other" because it gave us the freedom to put less-preferred (but still up-to-date) languages there.

How would we define "preferred"? In terms of the IOI and camp the only language we prefer is C++. The fact that we all like Python is favouritism on our part

puqeko: To throw another idea into the pop, implement a time relevant frequency counter that counts the number of submissions in each language upon submission. We would then lookup each language by frequency counter value and display then in that order. One issue here, aside from implementation, could be that the most popular language will oscillate on a yearly basis (Python by the end of NZIC and C++ by the end of camp).

tom93: > How would we define "preferred"?

I was thinking it would be to do with what we teach in camp (mainly C++, but the NZIC editorials use Python) but also what students commonly use.

> lookup each language by frequency counter

A hard-coded list of favourite languages is like a cheap version of this; we can manually run some queries on the database if we want to see which languages are popular, but then we can also factor in the camp's bias ("C++ first").

puqeko: Would it make sense then to have C++ and Python as the NZIC official languages and use three groups in the menu...

Official
    C++
    Python

Current
    [list alphabetically common and up-to-date languages]

Other
    [Old versions + random languages]

tom93: How about we merge the last two?
```
Favourite/Pinned/Official
    C++
    Python
Other
    [list everything alphabetically]
```

tom93: (if we can't settle on good names, we can use a separator, "----")

Holmes98: I think we couldn't work out how to add a non-selectable separator (without optgroups) but otherwise I like the last option the most.

puqeko: Agreed. The non-selectable separator was before I started using groups, so hopefully it shouldn't be difficult.

tom93: This makes a non-selectable separator: `["────────", "", {disabled: true}]`
(credit for the unicode magic: https://stackoverflow.com/questions/899148/html-select-option-separator)

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
 chroot "$ISOLATE_ROOT" update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 75
 # gcc 9 done
 
-# let user know that chroot installs are finished
+[ -z "$TRAVIS" ] && bash script/confirm.bash 'Install .Net Core (C#)' && {
```
tom93: The official capitalization is ".NET Core"

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  set -x
+
+  : Installing .Net Core for C#
+  # https://docs.microsoft.com/en-us/dotnet/core/install/linux-package-manager-ubuntu-1604
+
+  : Downloading package info from Microsoft
+  UBUNTU_VERSION=$(lsb_release --release --short)
```
tom93: This should be executed inside chroot (also the dotnet command when getting the version below).

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  : Creating runtime config file
+  # The dotnet command will not execute the compiled .exe unless there is a config file
+  # which tells it what runtime framework to use.
+  # https://stackoverflow.com/questions/46065777/is-it-possible-to-compile-a-single-c-sharp-code-file-with-the-net-core-roslyn-c
+
+  RUNTIME_CONFIG_PATH="$ISOLATE_ROOT/usr/share/dotnet-config.json"  # used in the compile command
+  RUNTIME_CONFIG_TEMPL="script/csharp/dotnet-config-template.json"
 
+  # referenced in RUNTIME_CONFIG_TEMPL file
+  DOTNET_FW_VERSION=$(dotnet --version | sed -r "s/([0-9].[0-9].[0-9])[0-9]*/\1/") \
```
tom93: We need to move the line that unmounts /proc to the end of the script, otherwise dotnet fails:
```
Failed to resolve full path of the current executable [/proc/self/exe]
```

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  set -x
+
+  : Installing .Net Core for C#
+  # https://docs.microsoft.com/en-us/dotnet/core/install/linux-package-manager-ubuntu-1604
+
+  : Downloading package info from Microsoft
+  UBUNTU_VERSION=$(lsb_release --release --short)
+  chroot "$ISOLATE_ROOT" wget -q "https://packages.microsoft.com/config/ubuntu/${UBUNTU_VERSION}/packages-microsoft-prod.deb" -O packages-microsoft-prod.deb
+  chroot "$ISOLATE_ROOT" dpkg -i packages-microsoft-prod.deb
+  chroot "$ISOLATE_ROOT" rm packages-microsoft-prod.deb
```
tom93: I would put the temporary file in /tmp rather than /.

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  [ "$UBUNTU_VERSION" == "18.04" ] && chroot "$ISOLATE_ROOT" sudo add-apt-repository universe
```
tom93: According do the [code](https://github.com/dotnet/core/commit/adc8888f3ab2) the 'universe' repository is always required, but only mentioned for Ubuntu 18.04 because in the others it's enabled by default. We might as well enable it unconditionally. (Also, no need for sudo.)

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  chroot "$ISOLATE_ROOT" apt-get install -y apt-transport-https
+  chroot "$ISOLATE_ROOT" apt-get update
+
+  : Installing dotnet 3.1
+  chroot "$ISOLATE_ROOT" apt-get install -y dotnet-sdk-3.1
+
+  : Creating runtime config file
+  # The dotnet command will not execute the compiled .exe unless there is a config file
+  # which tells it what runtime framework to use.
```
tom93: I would write this as
```
  # The dotnet command requires executables to have a corresponding file
  # [appname].runtimeconfig.json that specifies which runtime framework to use.
  # We create such a file and set it using the --runtimeconfig option.
```

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  # https://stackoverflow.com/questions/46065777/is-it-possible-to-compile-a-single-c-sharp-code-file-with-the-net-core-roslyn-c
+
+  RUNTIME_CONFIG_PATH="$ISOLATE_ROOT/usr/share/dotnet-config.json"  # used in the compile command
```
tom93: It's actually used in the interpreter command

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  RUNTIME_CONFIG_TEMPL="script/csharp/dotnet-config-template.json"
```
tom93: I don't think the variable $RUNTIME_CONFIG_TEMPL is helpful, I would inline the path into the envsubst command

tom93: I would rename the template to runtimeconfig-template.json and the destination to dotnet-runtimeconfig.json to try to match the standard name [appname].runtimeconfig.json (and rename the variable to $RUNTIMECONFIG_PATH).
I'm not completely happy with my proposed names, also considering template.runtimeconfig.json for the template and standalone.runtimeconfig.json or nztrain.runtimeconfig.json for the destination. (The name runtimeconfig.template.json already has a meaning in .NET Core projects.)

script/install/debootstrap.bash
tom93: This is trying to get the .NET Core Runtime version from the .NET Core SDK version, right? There are examples in https://docs.microsoft.com/en-us/dotnet/core/versions/ where the third components don't match. I think the correct way to get the runtime version would be to parse the output of [dotnet --list-runtimes](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet?tabs=netcore21), however the [Runtime Binding Behaviour design doc](https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md) recommends setting the patch number in runtimeconfig.json to 0 (which is much easier!).
```
  # Compute the .NET Core Runtime version from the .NET Core SDK version
  # The major and minor versions are the same (https://docs.microsoft.com/en-us/dotnet/core/versions/).
  # The patch version should be set to 0 (https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md).
  DOTNET_FW_VERSION=$(chroot "$ISOLATE_ROOT" dotnet --version | sed 's/\.[0-9]\+$/.0/') \
    ...
```
(If we wanted to simplify things further we could be sneaky and use the runtimeconfig of dotnet.dll itself, /usr/share/dotnet/sdk/3.1.101/dotnet.runtimeconfig.json, but that feels fragile.)

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
 
+  # referenced in RUNTIME_CONFIG_TEMPL file
+  DOTNET_FW_VERSION=$(dotnet --version | sed -r "s/([0-9].[0-9].[0-9])[0-9]*/\1/") \
+  DOTNET_FW_NAME="Microsoft.NETCore.App" \
+  envsubst < $RUNTIME_CONFIG_TEMPL | cat > $RUNTIME_CONFIG_PATH
```
tom93: There is no need for `cat`, we can use `envsubst < template > output` (I would also quote the variables and indent the continuation lines)

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+
+  : Moving shell scripts into place
```
tom93: It's actually copying

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  
+  cp "script/csharp/compile-command.sh" "$ISOLATE_ROOT/usr/local/csc.sh"
+  chroot "$ISOLATE_ROOT" ln "/usr/local/csc.sh" "/usr/bin/csc"
```
tom93: Any reason to both copy csc.sh _and_ create a symlink? It would make sense if we were using update-alternatives to manage these links, but we're not. I put it in /usr/local/bin, and copy it directly as /usr/local/bin/csc (the .sh suffix isn't significant).

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  chroot "$ISOLATE_ROOT" chmod +x "/usr/bin/csc"
+
+  cp "script/csharp/run-command.sh" "$ISOLATE_ROOT/usr/local/csr.sh"
```
tom93: I would rename the scripts to dotnet-csc and dotnet-exec (inspired by the mono-devel package, which installs mono-csc). I would also move the directory `scripts/csharp` into `scripts/install/csharp`.

script/install/debootstrap.bash
```diff
@@ -173,5 +173,47 @@ echo "$chroot_cmd update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9
...
+  # referenced in RUNTIME_CONFIG_TEMPL file
+  DOTNET_FW_VERSION=$(dotnet --version | sed -r "s/([0-9].[0-9].[0-9])[0-9]*/\1/") \
+  DOTNET_FW_NAME="Microsoft.NETCore.App" \
+  envsubst < $RUNTIME_CONFIG_TEMPL | cat > $RUNTIME_CONFIG_PATH
+
+  : Moving shell scripts into place
+  
+  cp "script/csharp/compile-command.sh" "$ISOLATE_ROOT/usr/local/csc.sh"
+  chroot "$ISOLATE_ROOT" ln "/usr/local/csc.sh" "/usr/bin/csc"
+  chroot "$ISOLATE_ROOT" chmod +x "/usr/bin/csc"
```
tom93: We can mark it as executable in version control, then we won't need chmod because cp will preserve the execute permission. It also makes it slightly easier to run the script directly from the version control directory.
(This might not work well on Windows, which doesn't have the notion of execute permission, but I don't think we are trying to support Windows.)

script/csharp/compile-command.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
puqeko: It does

script/csharp/compile-command.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# This script does two things
+#   1. Redirects all output to stderr
+#   2. Supplies .dll references to the csc command
+
+# Actually compile:
+# [dotnet] [csc] -reference:[stdlib.dll] [...passed in args...]
+VERSION=$(dotnet --version)
+/usr/bin/dotnet "/usr/share/dotnet/sdk/"$VERSION"/Roslyn/bincore/csc.dll" \
+-reference:"/usr/share/dotnet/sdk/"$VERSION"/ref/netstandard.dll" "$@" 1>&2
```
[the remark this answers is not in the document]
puqeko: > No need for full path to dotnet

Why then have we been specifying `/usr/bin/g++` etc for the rest of them? It would be more readable to keep it short. I tested this with g++ and it worked but python did not work without `/usr/bin`.

> csc.dll prints version number of the C# compiler, I think the .NET Core SDK version should be added

Makes sense. Do we bother removing the copyright line in keeping with GHC and GCC?

script/csharp/dotnet-config-template.json
```diff
@@ -0,0 +1,8 @@
+{
+    "runtimeOptions": {
+        "framework": {
+            "name": "$DOTNET_FW_NAME",
```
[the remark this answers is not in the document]
puqeko: Fair

script/install/debootstrap.bash
puqeko: Good catch

script/install/debootstrap.bash
puqeko: That's nicer

script/install/debootstrap.bash
puqeko: Good catch

script/install/debootstrap.bash
puqeko: I'd prefer to use `dotnet-runtimeconfig-template.json` with `dotnet-runtimeconfig.json`

script/install/debootstrap.bash
puqeko: I figured it probably wasn't reliable to assume the output of `dotnet --list-runtimes` will be consistent across versions. Good find with the 0 patch version

script/install/debootstrap.bash
puqeko: The intension was to be consistent with the other compile commands using `/usr/bin/...` since that part of the command displays on the server. Hence, I copied it into the sandbox and symlinked them in. Could we go with update-alternatives instead?

script/install/debootstrap.bash
puqeko: Nice

script/install/debootstrap.bash
puqeko: Cool, I didn't know that

puqeko:
Other than the few things I've pointed out, your changes seem reasonable to me

script/csharp/compile-command.sh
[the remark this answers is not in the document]
tom93: > Why then have we been specifying `/usr/bin/g++` etc for the rest of them? It would be more readable to keep it short. I tested this with g++ and it worked but python did not work without `/usr/bin`. Perhaps further investigation in another PR.

When a command (array of arguments) is executed using isolate, the first argument needs to be the full path to the program and the remaining arguments are passed as argv[1] etc. At first, the compilation commands were executed directly using isolate so the first argument needed to be the full path to the compiler (e.g. `/usr/bin/g++`).

When Java was added (32dcf29af8ed249792e73d744af97f29d16d8b00) it required two commands during compilation, javac and jar, so the compilation commands were wrapped in bash (e.g. first arg = `/bin/bash`, second arg = `-c`, third arg = `javac ... && jar ...`). Since the compilation command is now executed by bash, it can use shell operators like &&. It also doesn't need to use full paths to programs because bash searches for them in $PATH.

Python still requires a full path is because it uses an interpreter, and the interpreter commands aren't wrapped in bash.

All this is only for commands in db/languages.yml that are going to be executed directly by isolate, i.e. interpreter commands and (historically) compilation commands. Inside a bash script like `compile-command.sh` there is no need to use full paths to programs because bash searches in $PATH.

> > csc.dll prints version number of the C# compiler, I think the .NET Core SDK version should be added
> 
> Makes sense. Do we bother removing the copyright line in keeping with GHC and GCC?

There is an option `-nologo` that removes the copyright message, so sure.

script/install/debootstrap.bash
tom93: Sure

script/csharp/compile-command.sh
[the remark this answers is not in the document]
puqeko: Ah okay. Makes sense. In which case, I'd like to propose that we remove `/usr/bin` from all the current compile commands to make them more readable to users. I'll open another PR

puqeko:
Oops. I’ll probably also look into changing the commit signature before we merge this (if that’s possible)

tom93:
I'd prefer to use squash-and-merge because the history is complicated and not interesting. So there will be one commit with a new commit message (we can add "Co-Authored-By:" lines if we want). If the squash-and-merge is done through GitHub, it sets the commit author to the author of the PR and the committer to "GitHub \<noreply<span>@</span>github.com\>". (Not sure what you mean by commit signature.)

puqeko:
I made 6b95210 before I saw your message. Is that sort-of inline with what you were thinking? By signature I meant that it had my work email which probably isn't a great idea to have

tom93:
Yes, it's exactly what I meant.
(The squash introduced some changes, e.g. the execute permission on the scripts was lost -- probably a Windows thing -- but I'm planning to add more review comments, so we could re-squash later.)

puqeko:
Okay, sounds good. Yeah I'm on a windows machine at work currently. Filling in time while stuff installs

puqeko:
Should I force push?

tom93:
I'd like to keep the full history in this pull request (and there are undesirable differences in the squashed commit).

tom93:
As discussed in chat, running `dotnet` using isolate on Linux < 4.14 produces the following error:
```
Failed to create CoreCLR, HRESULT: 0x8007001F
```
Running `strace dotnet` shows that it tries to use mlock, which isolate prevents:
```
...
membarrier(MEMBARRIER_CMD_QUERY, 0)     = 0x1 (MEMBARRIER_CMD_SHARED)
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x2b82ffccf000
mlock(0x2b82ffccf000, 4096)             = -1 EPERM (Operation not permitted)
write(2, "Failed to create CoreCLR, HRESUL"..., 45Failed to create CoreCLR, HRESULT: 0x8007001F) = 45
```

It works on Linux >= 4.14 because then MEMBARRIER_CMD_PRIVATE_EXPEDITED is available, and dotnet uses that instead of mlock (see [https://github.com/dotnet/runtime/issues/10568](https://r.duckduckgo.com/l/?uddg=https%3A%2F%2Fgithub.com%2Fdotnet%2Fruntime%2Fissues%2F10568), [https://github.com/dotnet/coreclr/pull/20949](https://r.duckduckgo.com/l/?uddg=https%3A%2F%2Fgithub.com%2Fdotnet%2Fcoreclr%2Fpull%2F20949)).

The server currently runs Linux 4.4.0, so we need to update the kernel before deploying this PR. I will create an issue to track that.

script/install/csharp/dotnet-csc
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# This script does two things
+#   1. Redirects all output to stderr
+#   2. Supplies .dll references to the csc command
+
+# Actually compile:
+# [dotnet] [csc] -reference:[stdlib.dll] [...passed in args...]
```
tom93: Does this intentionally say "stdlib.dll" instead of "netstandard.dll"?
I like "stdlib" (without .dll) because it makes more sense to me than "netstandard.dll", but with .dll it sounds like it's referring to a specific file rather than describing the purpose of a file.

puqeko: The "stdlib" is intentional but I don't think the ".dll" is. I agree with "stdlib" only.

puqeko:
I haven't tested them but those changes seem reasonable to me. Making the quoting consistent for other languages could become part of PR #70 since that's related cleaning up `langauges.yml`. Telemetry isn't any help to us. Should we re-open this PR so that we can see it's still active under the PR tab?

Holmes98:
Deployed successfully.

PR #65, opened 2020-01-23
Make J's version number consistent

puqeko:
2020-01-23 at 11 17 03 PM

### DIFF
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -136,5 +136,5 @@ j:
   :interpreter_command: '%{interpreter} %{source}'
   :variants:
     j8:
-      :name: "J 8.0.3"
+      :name: "J 8.0"
       :interpreter: /usr/bin/ijconsole-8.0.3
--- a/db/languages.yml
+++ b/db/languages.yml
@@ -136,5 +136,5 @@ j:
   :interpreter_command: '%{interpreter} %{source}'
   :variants:
     j8:
-      :name: "J 8.0"
+      :name: "J 8.03"
       :interpreter: /usr/bin/ijconsole-8.0.3


Holmes98:
Hmm. To be honest, I kind of prefer J to have the full version number since its name is already short enough. What does everyone else think?

puqeko:
My reasoning was for consistency with `Python 3.8`, `Ruby 2.2`, and possibly the soon to be `Javascript (V8 8.1)` and `C# 8.0`.

puqeko:
I guess a semi-related topic is weather `C++ 17` and `C 11` should have spaces, but it was the `J` thing that was annoying me most though.

Holmes98:
I think we should be consistent with whatever the official documentation uses.
- [C++17 and C11 should definitely not have spaces](https://isocpp.org/std/the-standard).
- [J's wiki uses J803](https://code.jsoftware.com/wiki/System/ReleaseNotes/J803) though I think that could be confusing. Based on the version numbering from [other pages](https://code.jsoftware.com/wiki/System/ReleaseNotes/J806) I think `J 8.03` would be more appropriate.

db/languages.yml
```diff
@@ -136,5 +136,5 @@ j:
   :interpreter_command: '%{interpreter} %{source}'
   :variants:
     j8:
-      :name: "J 8.0.3"
+      :name: "J 8.0"
```
Holmes98:
```suggestion
      :name: "J 8.03"
```

puqeko: Hmmm. Going by offical docs seems a good benchmark. They use a weird numbering system. `J803` would be the proper way to do it as you say. However, if I google `J803` the first result is for Java whereas if I google `J 8.03` I get the J (programming language) wikipedia page, which is much more reasonable. So, although this wasn't the change I was hoping for with this PR, I agree with you

PR #67, opened 2020-01-29
Set each user's default option to their last submitted lanugage

Holmes98:
See https://github.com/NZOI/nztrain/pull/64#discussion_r372315509.

If the last submitted language is no longer selectable, or the user has not yet made any submissions, it will default to the first option in the dropdown. Also, since the ordering of the current languages
is manually defined, I've removed `Language.default`.

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -59,7 +59,7 @@ class Language < ActiveRecord::Base
   # for the selection dropdown
   def self.grouped_submission_options
     current = ["c++", "c", "python", "java", "haskell", "ruby", "j"]  # language group identifiers, order is preserved
-    other = ["c++14", "c++11", "c99"]  # language identifiers, ordered by language name
+    other = ["c++14", "c99"]  # language identifiers, ordered by language name
     current_langs = LanguageGroup.where(:identifier => current).preload(:current_language).map(&:current_language)
     current_options = current_langs.sort_by { |lang| current.index(lang.group.identifier) }.map { |lang| [lang.name, lang.id] }
     other_options = Language.where(:identifier => other).order(:name).pluck(:name, :id)
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -67,10 +67,6 @@ class Language < ActiveRecord::Base
       "Other" => Hash[other_options] }
   end
 
-  def self.default
-    LanguageGroup.find_by(identifier: 'c++').current_language
-  end
-
   def self.infer(ext)
     case ext
     when *%w[.cpp]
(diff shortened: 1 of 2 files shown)


puqeko:
> I'll later submit a PR that will set each user's default language to the language of their latest submission.

Sounds great. The issue of not noticing when a newer version becomes available is annoying though, and will come up again every time we update the language.

Would it be too complicated/slow to add some logic along the lines of

"If the last submission's language is no longer the current version within its group, and the user has never submitted in the current version of the language, then use the current version of that language as the default."

This is meant to catch cases where a user tries C++17 but then intentionally wants to go back to C++14. But if a user is careful to never submit in C++17, it will keep trying to set it as their default. The only solution I can think of is to also record the date when the language version was added, which seems too much hassle.

> Perhaps a better name for the "Other" group is "Older" or "Outdated"? In which case, Ruby fits (for now).

I thought that at first, but I started liking "Other" because it gave us the freedom to put less-preferred (but still up-to-date) languages there. So we could make C++ and Python into the "Current" languages, and everything else into "Other". (But then maybe rename "Current" to "Favourites" or something.)

_Originally posted by @tom93 in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjMwMTg3ODM1OnYy/pull_request_review_threads/more_comments_

puqeko:
I think another suggestion from @Holmes98  was to have a "news" or "change log" on the front page to make users aware of language updates?

Holmes98:
> Would it be too complicated/slow to add some logic along the lines of
> 
> "If the last submission's language is no longer the current version within its group, and the user has never submitted in the current version of the language, then use the current version of that language as the default."
> 
> This is meant to catch cases where a user tries C++17 but then intentionally wants to go back to C++14. But if a user is careful to never submit in C++17, it will keep trying to set it as their default. The only solution I can think of is to also record the date when the language version was added, which seems too much hassle.

This does seem rather difficult to implement. Since the current changes are quite small and it will likely be a while before we update languages again, I've just gone with my original idea for now. If we update Python I think we'd be likely to remove 3.8 from the submission options, so C++ seems to be the only relevant case here.

tom93:
> This does seem rather difficult to implement.

Yes. I like the current version, it's surprising how simple it was to grab the last submission's language.

Holmes98:
Oops, turns out the [`&.` operator](https://stackoverflow.com/questions/36812647/what-does-ampersand-dot-mean-in-ruby) was only introduced in Ruby 2.3, so this can't be deployed yet.

PR #69, opened 2020-01-30
Fix null attributes of timestamps to match server

tom93:
See commit messages for details.

This pull request changes the timestamp columns in the version-controlled schema and the migrations to match the current schema on the server. If we want to make all the timestamp columns consistently use `null: true` or `null: false` it should be done separately.

Note that there is another inconsistency which is not addressed here: the `limit: 128` attribute on the users.encrypted_passwords column (also introduced in [fe025a8](https://github.com/NZOI/nztrain/commit/fe025a81727e2e467b6f9f7ed7be4f8222c50709#diff-1acd2e7e27a227829d5d14a91c863bb6L517)).

### DIFF
--- a/db/migrate/20110819224757_devise_create_users.rb
+++ b/db/migrate/20110819224757_devise_create_users.rb
@@ -37,7 +37,7 @@ class DeviseCreateUsers < ActiveRecord::Migration
       # t.string :authentication_token
 
       t.boolean :is_admin, :default => false
-      t.timestamps
+      t.timestamps null: true
     end
 
     add_index :users, :email,                :unique => true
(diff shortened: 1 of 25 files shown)


tom93:
See commit messages for details.

This pull request changes the timestamp columns in the version-controlled schema and the migrations to match the current schema on the server. If we want to make all the timestamp columns consistently use `null: true` or `null: false` it should be done separately.

Note that there is another inconsistency which is not addressed here: the `limit: 128` attribute on the users.encrypted_passwords column (also introduced in [fe025a8](https://github.com/NZOI/nztrain/commit/fe025a81727e2e467b6f9f7ed7be4f8222c50709#diff-1acd2e7e27a227829d5d14a91c863bb6L517)).

PR #70, opened 2020-01-30
Simplify compilation commands

puqeko:


tom93:
I suggest squash-merging (after #98 is merged) with the following commit message:

```
Remove /usr/bin prefix from compile commands (#70)

Since commit 32dcf29 (add Java support, 2014-12-25), the compilation
commands are executed using bash instead of directly using isolate.
They no longer need the first argument to be the full path to the
compiler.

Note that the interpreter commands are still executed directly using
isolate and need to keep using the full path to the interpreter.

Co-authored-by: Tom Levy <tomlevy93@gmail.com>
```

puqeko:
Server restarted and database seeded. Updated commands verified by test submission.

PR #71, opened 2020-02-02
bundle list [args] is deprecated

puqeko:
Make the command actually do what it says it does.

### DIFF
--- a/script/install/whenever.bash
+++ b/script/install/whenever.bash
@@ -7,3 +7,4 @@ if [[ "$lastupdate" < "$lastmodified" ]] ; then
   echo '$ whenever --update-crontab '`basename \`pwd\``
   bundle exec whenever --update-crontab `basename \`pwd\``
 fi
+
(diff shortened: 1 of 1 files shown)


script/install/whenever.bash
```diff
@@ -5,6 +5,6 @@ lastupdate=`cat script/update.time` # get time of last update
 lastmodified=`find config/schedule.rb config/backup.yml -printf '%T+\n' | sort -r | head -n1`
 if [[ "$lastupdate" < "$lastmodified" ]] ; then
   echo '$ whenever --update-crontab '`basename \`pwd\``
-  `bundle list whenever`/bin/whenever --update-crontab `basename \`pwd\``
+  whenever --update-crontab `basename \`pwd\``
```
tom93:
```suggestion
  bundle exec whenever --update-crontab `basename \`pwd\``
```
This guarantees the correct version of 'whenever' runs, and is required on fresh installations that use rbenv (which only creates the shims on the next login, and only if configured in .bashrc).

puqeko: Sure

PR #74, opened 2020-02-10
Front-end improvements

Holmes98:


app/controllers/groups/members_controller.rb
```diff
@@ -31,6 +31,28 @@ def join
...
+  end
+
+  def remove
+    @group = Group.find(params[:id])
+    authorize @group, :remove?
+    @user = User.find(params[:user_id])
+    @group.members.delete(@user)
+    redirect_to(members_group_path(@group), :notice => "#{@user.username} has been removed from this group")
+  end
+
```
tom93: I think the new add/remove methods should be before the existing join method rather than between join/leave (unless there is a particular reason to put them there?)

app/policies/group_policy.rb
```diff
@@ -48,6 +48,14 @@ def invite?
...
   end
 
+  def add?
+    user.is_admin?
+  end
+
+  def remove?
+    user.is_admin?
+  end
+
```
tom93: Similar thing to the controller, I'd prefer add?/remove? to be before join? etc. rather than between invite?/reject?.
I also think the names should be add_user?/remove_user?, because groups can also have contests and problem sets added and removed.

app/policies/group_policy.rb
```diff
@@ -48,6 +48,14 @@ def invite?
     user.is_admin? or user.owns(record) or member? && [Group::MEMBERSHIP[:open], Group::MEMBERSHIP[:invitation]].include?(record.membership)
   end
 
+  def add?
+    user.is_admin?
+  end
+
+  def remove?
+    user.is_admin?
```
tom93: I wonder if we should also allow group owners to remove members (even if the group owner is not an admin)...

app/views/groups/members/index.html.erb
```diff
@@ -1,17 +1,31 @@
...
+      <% end %>
     </tr>
   </thead>
   <tbody>
     <% @memberships.each do |membership| %>
       <tr>
         <td><%= link_to membership.member.username, membership.member %></td>
         <td><%= membership.created_at %></td>
+        <% if policy(@group).remove? %>
+          <td><%= link_to("Remove", remove_members_group_path(@group, :user_id => membership.member.id ), :data => { :confirm => 'Are you sure?' }, :method => "put") %></td>
```
tom93:
```suggestion
          <td><%= link_to("Remove", remove_members_group_path(@group, :user_id => membership.member.id), :data => { :confirm => 'Are you sure?' }, :method => "put") %></td>
```

app/controllers/groups/members_controller.rb
Holmes98: Sure, I'm fine with that.

app/policies/group_policy.rb
Holmes98: I wouldn't mind that.

Holmes98:
Deployed.

PR #78, opened 2020-02-12
Show failing sample case details by default

### LINKED ISSUE #77: Make sample data details on submission page show by default
As suggested by @Holmes98 , to help students who are new to nztrain, we should automatically show input, output, and expected output for failing sample cases on the submission page.

Holmes98:
Closes #77.


app/views/submissions/show.html.erb
```diff
@@ -107,7 +107,7 @@
             <td></td>
           </tr>
           <% if ![:pending,:cancelled].include?(case_data.status) && (policy(@submission).inspect? || is_sample_set) %>
-            <tr class="inspect_test status_<%= case_data.status %>">
+            <tr class="inspect_test status_<%= case_data.status %><%= " sample_case" if is_sample_set %>">
```
tom93: I feel like the 'sample_case' class should logically be on the first <tr> element (that also has the 'test_case' class), rather than on the second <tr> element.

tom93:
Deployed

PR #79, opened 2020-02-14
Update README and add Dockerfile

Techno-coder:
This adds a `Dockerfile` for building under Docker. The `README.rdoc` has been updated to reflect building under Docker or manually.


Dockerfile
```diff
@@ -0,0 +1,67 @@
+# Build invocation:
+# May need to be executed with --memory-swap -1 or else out of memory error
+# docker build --memory-swap -1 . -t nztrain:latest
+
+# Execution invocation:
+# Privileged mode is required for control groups
+# Must be invoked with '--network host' for port exposure
+# docker run --privileged --network host -it nztrain:latest
```
bagedevimo, line 8: -p / --expose should also be suitable no?

Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN bundle install --deployment
+
+COPY . .
+RUN service postgresql start; \
+    service redis start; \
+    bash script/update.bash; \
+    rm /var/run/redis/redis.pid; \
+    service postgresql stop
+
+CMD service redis start; \
```
bagedevimo, line 63: IMO we are better to run redis/postgres outside this container and tell the app where to find them via env vars - in host and bridge (default iirc) - means this script can be vastly simplified and we can take advantsge oh GH actions having coomunity support for booting these services

PR #81, opened 2020-02-14
Use Docker for Travis CI build

Techno-coder:
This pull request depends on #79 and should be rebased. Initial builds will be slow as caches (Docker images) need to be built.


Techno-coder:
Unfeasible due to high overhead of cache creation by Travis CI.

PR #83, opened 2020-02-15
Fix scroll draw on Safari (deploy)

Techno-coder:
Cherry pick of #80.

### DIFF
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -113,7 +113,6 @@ a.progress_link:hover {
   right: 0;
   bottom: 0;
   overflow-y: auto;
-  outline: none;
 }
 
 #main-container
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -113,6 +113,7 @@ a.progress_link:hover {
   right: 0;
   bottom: 0;
   overflow-y: auto;
+  outline: none;
 }
 
 #main-container
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -109,10 +109,9 @@ a.progress_link:hover {
 
 #main {
   top: 30px;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/style.css
```diff
@@ -109,11 +109,11 @@ a.progress_link:hover {
 
 #main {
   top: 30px;
-  left: 200px;
+  margin-left: 200px;
   right: 0;
   bottom: 0;
-  position: fixed;
   overflow-y: auto;
+  outline: none;
```
puqeko: What is the function of adding `outline: none`?

Techno-coder: Hovering or clicking inside the main container (problem statement area) creates a "selected" outline.

puqeko: The default behaviour of outline should be none by default. This seems to be a consequence of the `tabindex="-1"` attribute. The only two reasons I can think of for needing `tabindex="-1"` is for accessibility and to facilitate automated tests. As far as accessibility, removing it doesn't seem to change the tab order of the page. If coveralls returns the same result, I would be inclined to remove the tabindex attribute instead

Techno-coder: Per mirrored commit on https://github.com/NZOI/nztrain/pull/80, coveralls remains unchanged.

puqeko:
Deployed and restarted

Note: the main window appears to have moved 1-2 pixels with no obvious consequences.

tom93:
This PR was reverted (by force-pushing the parent commit to the 'deployed' branch) because of an issue reported by @puqeko in chat.

> If you scroll up and down the menu bar stops sticking to the top and leaves a white bar (on macOS, using Opera, Safari, and Chrome but not Firefox).
> Some of that behaviour was preexisting but it was much more noticeable/pronounced after the change.

screenshot

PR #85, opened 2020-02-15
Fix menu bar fixed position

Techno-coder:


puqeko:
2020-02-15 at 9 56 29 PM

puqeko:
Seems to not be an issue in newer Safari versions. See #124

PR #90, opened 2020-02-22
Remove padding around headers in test case inspection table

puqeko:
The header elements inside the inspection table when viewing a submission test case result would make a new line when it was not necessary to do so.

Have decreased font size and removed padding.

Before:
2020-02-22 at 9 39 11 PM

After:
2020-02-22 at 10 12 16 PM


app/assets/stylesheets/submission.css
```diff
@@ -13,6 +13,11 @@ table.results {
   vertical-align: top;
 }
 
+.results .inspect_output th {
```
tom93: It belongs next to the existing rules for `.results .inspect_test table.inspect_output` further down (the selector should be updated to match).

app/assets/stylesheets/submission.css
```diff
@@ -13,6 +13,11 @@ table.results {
...
+  padding: 0;
```
tom93: The change breaks the alignment with "Wall time" from the previous row.
Before:
1-before
After:
2-after
It would be nice if they remained aligned, but I don't have a good way of doing that (yet).

app/assets/stylesheets/submission.css
```diff
@@ -13,6 +13,11 @@ table.results {
...
+  font-size: 0.95em;
```
tom93: I don't think we should reduce the font size. Word wrapping is inevitable on certain display sizes, and reducing the font size is not a sustainable solution.

app/assets/stylesheets/submission.css
```diff
@@ -13,6 +13,11 @@ table.results {
...
+}
+
```
tom93: Currently the headings (e.g. "Input") and the stats (e.g. "Size: 2 Bytes") are in separate columns of equal width, which can cause unnecessary word wrapping since each must fit within its own half.
How about we merge the columns and use an alternative way to right-align the stats, for example a \<span\> element with `float:right`? It changes the behaviour when wrapping, but to me it looks fine:
3-float-right

puqeko: Yeah, I like this idea

app/assets/stylesheets/submission.css
puqeko: I don't think this is a major concern since the wall time information isn't available to standard users. But yes, if there is an easy fix it would be nice

Holmes98: Also note that the Wall time is already slightly misaligned in the original version (due to `border-spacing: 5px;`)

app/assets/stylesheets/submission.css
Holmes98: I'd also prefer keeping it at the standard font size.

app/assets/stylesheets/submission.css
puqeko: How about?
2020-02-24 at 12 07 39 AM

puqeko:
Thoughts on this layout?
2020-02-24 at 12 12 20 AM

Holmes98:
> Thoughts on this layout?

 - I'd prefer the times to be not bold. `Time exited` seems redundant.
 - I'd prefer the inspection table to just have 4 columns rather than 2 + 3, as it makes it easier to view when browsing through the test cases. I think the sandbox message usually isn't important and was fine next to the wall time.
- The input/actual output sizes are too close to the next heading on the right. In this respect, I actually kind of prefer the current design (with padding on the left+right), as it gave more of a separation between the two, though it could definitely be improved.

tom93:
> * I'd prefer the inspection table to just have 4 columns rather than 2 + 3, as it makes it easier to view when browsing through the test cases. I think the sandbox message usually isn't important and was fine next to the wall time.

I also prefer 4 columns, it works well enough on a standard monitor size.
We could hide the 'Evaluator Log' column when it says 'status:OK' (i.e. for problems without a custom evaluator).
 
> * The input/actual output sizes are too close to the next heading on the right. In this respect, I actually kind of prefer the current design (with padding on the left+right), as it gave more of a separation between the two, though it could definitely be improved.

I agree that they are too close when there is no padding at all. The earlier 'After' screenshots have a 10px padding from [here](https://github.com/NZOI/nztrain/blob/fc59e66bd30c1e0261ff23d2d7d6dbc3988c7b98/app/assets/stylesheets/submission.css#L140), do you think that's enough?

@puqeko Assuming we merge the heading/stat columns with `float:right`, how do you feel about the padding and word wrap? For example, would it be enough to reduce the horizontal padding from 20px to 10px? Or even keep it at 20px?

Holmes98:
> The earlier 'After' screenshots have a 10px padding from here, do you think that's enough?

I'd prefer them to be separated more, maybe 20px?

tom93:
For reference, here it is with float:right and the original padding (20px on left, 10px on right):
1-float-right-20px-10px
And with 10px padding:
2-float-right-10px-10px
(I also reduced the padding on the entire inspect_test)
Code in branch 'fix-inspection-formatting-float-right' (abcbed3253d8028e09ca4bec7e00eb83eb97d136).

puqeko:
- Removed access whitespace between header info at top of submission.
- Aligned admin inspect info headers `Time (killed):` to `Input` header in inspection window.
- Used 10px right padding for floated size labels (can increase to 20 px if prefered)

2020-04-19 at 5 08 27 PM

puqeko:
@tom93 Are you good with merging this?

Holmes98 (changes requested):
I've made some changes to Tom's version under `fix-inspection-formatting-float-right-v2`, and would be happy to squash-merge that in it's current state.

With regards to this PR:
- I'm unsure about removing the spacing between the problem/language/score etc. fields. I do like how it's more consistent with the contest pages, but for now I'd prefer to leave it as is.
- It's worth noting that hiding the evaluator log also removes the evaluation score. This could still be useful in some cases, even when the problem does not use a custom evaluator – for example, see https://train.nzoi.org.nz/submissions/56111 (it has runtime errors, but they still printed the correct output). For now, I've left this out of the proposed branch.

puqeko:
The motivations for removing the spacing were A: for consistency and B: to take up less space above the submission breakdown.

I still don't quite understand what the benefit is of keeping the evaluator log unless the message is non-standard. Even if the evaluation score is not 0 or 1 (in the rare case of a custom evaluator that uses this feature) then the effect of the evaluator score will still show in the main table. In the example given, it is already clear that a runtime error occurred. The evaluator log doesn't provide any further information that is useful for debugging as far as I can tell? It just seems like unnecessary and confusing clutter to me.

Holmes98:
> The motivations for removing the spacing were A: for consistency and B: to take up less space above the submission breakdown.

It just looks a bit off to me, perhaps because I'm used to the old layout. But I don't really mind, so feel free to change it under `fix-inspection-formatting-float-right-v2`.

> The evaluator log doesn't provide any further information that is useful for debugging as far as I can tell?

The evaluator log doesn't provide further information, but the evaluation score does, in the case of a runtime error. In that particular example it's obvious that the outputs are identical, but there may be other cases where the outputs are non-trivial.

The log is only visible to staff/problem owners, and I've never really had an issue with horizontal spacing so I don't see any issue in leaving it. If you feel strongly about hiding the log then I'd prefer to move the evaluation score somewhere else to keep it visible.

puqeko:
Replaced by #131

PR #101, opened 2020-03-02
Update RSpec from 2 to 3.9

tom93:
This is required to support rake 11.x for #100, see https://github.com/rspec/rspec-core/commit/8e723fc805e901ac4fa5483837138b175d411d6e.


tom93:
I adopted Holmes98's commit 73e6fc2 (Convert specs to RSpec 3.9.1 syntax with Transpec) in favour of 4dbcc40 (Explicitly enable rspec's old `should` syntax to dismiss deprecation warning), and reordered the commits slightly so Transpec does more of the work (6f2aeec reduces to e00dfe5).
For ability_spec_helper, I completely rewrote it as a custom matcher (8dd4a34) instead of including RSpec::Matchers in the user class (e8a8022).
I also removed the include of rspec/autorun, which is deprecated (9f61389).

bagedevimo (dismissed):
👍

Holmes98 (approved):

PR #105, opened 2020-04-19
Nice messages

puqeko:
Open for discussion, but I always disliked the submission messages. Particularly "There's room for improvement". Would be nice to use things which are more encouraging.

2020-04-19 at 10 28 42 PM

### DIFF
--- a/app/views/submissions/show.html.erb
+++ b/app/views/submissions/show.html.erb
@@ -11,13 +11,17 @@
   </p>
 <% else %>
   <div id="submission_judged" class="score_<%= (@submission.score/10).to_i %>x">
-    <%= @submission.score %>% - <%= case
-                                   when @submission.score == 100; "All done. Nice."
-                                   when @submission.score >= 80; "Close, but not quite."
-                                   when @submission.score >= 40; "There's room for improvement"
-                                   else; "Sorry, try again."
-                                   end
-                               %>
+    <%= @submission.score %>% - <%=
+    case
+      when @submission.score == 100; ["Top marks!", "One hundo!", "Smashed it!", "Phenomenal.", "Like a boss.", "Top stuff.", "Touch down!"].sample
+      when @submission.score >= 90; ["Fantastic! Soooooooo close!", "So close!"].sample
+      when @submission.score >= 75; ["Great! Keep working at it.", "Nice one! Keep pondering."].sample
+      when @submission.score >= 50; ["Good work! Keep thinking.", "Solid.", "Keep it up!"].sample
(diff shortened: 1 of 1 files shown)


puqeko:
Could someone please review this?

Holmes98 (commented):
I agree with this idea in general, though I feel like the messages between scores 25-89 are too similar. Actually, the `>= 25` messages (which don't have any indication that you still need to do more work) sound like you've scored better than the `>= 75` messages. Maybe combine the 25, 50, 75 cases and make the messages more consistent?

I'm also unsure about if it'd be better to stick to a single message under each case, so the same message always appears for the same score. Otherwise, `sample` should take an `rng` argument, so the message doesn't change if you refresh the page.

puqeko:
> I agree with this idea in general, though I feel like the messages between scores 25-89 are too similar. Actually, the `>= 25` messages (which don't have any indication that you still need to do more work) sound like you've scored better than the `>= 75` messages. Maybe combine the 25, 50, 75 cases and make the messages more consistent?
I'm okay combining the 50 and 75 cases. In the case of the `>= 25` messages, I see where you are coming from. My intent was to compensate for the large demoralising red/orange backgrounds and low percentage scores with a positive message. Hence, I didn't think it was worth pointing out the obvious that they need to do more work to get 100%. If we are trying to encourage students to do subtasks then I think we should try to be mindful of not putting them down for low scores. However, no good in over compensating (possibly causing confusion) either. Another approach could be to change the colours to green/yellow for all non-zero scores?

> I'm also unsure about if it'd be better to stick to a single message under each case, so the same message always appears for the same score. Otherwise, `sample` should take an `rng` argument, so the message doesn't change if you refresh the page.
I don't quite see why the message needs to be consistent every time the page is refreshed. Could you please clarify re your thinking on that? I think my intent was to introduce a bit of randomness and variety to make the low-score judgement seem less cold

eric-ide:
Just to add my 2 cents:
I agree with @Holmes98 that the message shouldn't change when you refresh it. If you're looking at a submission and refresh (or some back to it), and the message suddenly changes, it could be confusing and un-intuitive. Even though the score is unchanged, you would expect to see the exact same page when you refresh it.
However, I think it would be nice to have a bit of variety in the messages _per submission_, so I like the suggestion of
> sample should take an rng argument, so the message doesn't change if you refresh the page.

puqeko:
Okay. Thanks for looking at it. I don't agree with you guys but will got with Majority.

I don't know Ruby and I don't have a sandbox to test my changes but I presume `Random.new(@submission.id)` for the Random Number Generator (rng) argument to `sample`. Hopefully that works

Holmes98 (approved):

Issue #37, opened 2019-03-26
Investigate geocoder dependency


lambda-fairy:
NZTrain depends on a geocoder library.

What does it do? Do we still need it?

tom93:
The GeoIP database installation has recently broken, causing this to be revisited in issue #88 and PR #108.

tom93:
Removed in #108.

PR #108, opened 2020-04-22
Remove MaxMind GeoIP database and related code

### LINKED ISSUE #88: MaxMind GeoIP database requires license, breaks sign up on fresh installs
**Errors**

During installation, the `geoipupdate` command fails with one of the following error messages (depending on its version):
```
Received an unexpected HTTP status code of 401 from https://updates.maxmind.com/app/update_secure?db_md5=00000000000000000000000000000000&challenge_md5=3a5d812fa6d536c0e18a6271023f5669&user_id=999999&edition_id=GeoLite2-City
```
```
error retrieving updates: unexpected HTTP status code: 401 Unauthorized: An account ID and license key are required to use this service.
```

The failure causes the registration page (/accounts/sign_up) to crash with the following error:
```
IOError - GeoIP2 - Error opening the specified MaxMind DB file: /usr/share/GeoIP/GeoLite2-City.mmdb:
  app/views/accounts/registrations/new.html.erb:22:in `block in _app_views_accounts_registrations_new_html_erb__...'
  app/views/accounts/registrations/new.html.erb:3:in `_app_views_accounts_registrations_new_html_erb__...'
  lib/rack/response_timer.rb:12:in `_call'
  lib/rack/response_timer.rb:7:in `call'
```

**Cause**

The MaxMind GeoLite2 database requires a (free) license key since 30 December 2019 due to issues related to privacy regulations. See https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/ for details.

**Workaround**

The sign up page can be fixed by changing https://github.com/NZOI/nztrain/blob/97d96ac3002ce39ca57a3bd26e3bcf760f7bad78/app/views/accounts/registrations/new.html.erb#L22 to
```ruby
  <%= f.country_select(:country_code, selected: @user.country_code, priority_countries: ["NZ", "AU"]) %></p>
```

**Proposed Solution**

I propose removing the GeoIP database altogether. As far as I can tell, its only purpose is to set the default country on the registration page. To me it's not worth the hassle of obtaining a license key (and maintaining the install script, and downloading the packages).

Since almost all our users are from New Zealand, I suggest we simply set the default to NZ. If we don't want non-NZ users to accidentally leave the country as NZ, we can set the default to a placeholder and force users to explicitly select a country (the cost is that the majority of the users will then need two additional clicks to select NZ).

**Alternative Solution**
[Replace with a different database](https://github.com/NZOI/nztrain/pull/108#issuecomment-618150282)

**Related issues**
#37 - Investigate geocoder dependency

tom93:
The MaxMind GeoLite2 database requires a (free) license key since 30 December 2019 due to issues related to privacy regulations. See [1] and issue #88 for details.

The GeoIP database is only used to set the default country on the registration page and in an obsolete database migration that guesses the country of users who registered before the country field was added to the registration form.

I don't think it's worth the hassle of obtaining a license key and maintaining the install script.

If this PR is merged, New Zealand will become the default country on the registration page for all users.

(Side note: The maxmind.bash script also has a small bug. It uses add-apt-repository before installing software-properties-common.)

Closes #88 and #37.

[1] https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/


tom93:
Uninstall commands:
```
gem uninstall geocoder hive_geoip2
sudo rm /etc/GeoIP.conf # backup first
sudo apt-get purge geoipupdate
sudo apt-get purge libmaxminddb0 libmaxminddb-dev mmdb-bin
sudo add-apt-repository --remove ppa:maxmind/ppa
# optional: clean up /etc/apt/sources.list.d/maxmind-ubuntu-ppa-xenial.list (should be empty) and /etc/apt/trusted.gpg.d/maxmind_ubuntu_ppa.gpg
```

tom93:
**Alternative:** Replace MaxMind GeoLite2 database and keep geocoder

If we wanted to keep the geolocation feature, we could replace the MaxMind database with a different one.

For example, Ubuntu installs the "geoip-database" package by default, which provides /usr/share/GeoIP/GeoIP.dat. It's in the original GeoLite format (not GeoLite2), so to use it we would need to:

  - Update config/initializers/geocoder.rb to use [maxmind_local](https://github.com/alexreisner/geocoder/blob/master/README_API_GUIDE.md#maxmind-local-maxmind_local---experimental):

    ```
    Geocoder.configure(
      ip_lookup: :maxmind_local,
      maxmind_local: {
        file: File.join('/usr/share/GeoIP/', 'GeoIP.dat')
      }
    )
    ```

  - Add the geoip gem (instead of hive_geoip2):
    `bundle add geoip`

  - Patch geocoder:
    In [lib/geocoder/lookups/maxmind_local.rb:33](https://github.com/alexreisner/geocoder/blob/5ce27f03c88c968e15adc2fbd0e2fad65d319d53/lib/geocoder/lookups/maxmind_local.rb#L33), change `.city(...)` to `.country(...)`.
    Without this, it fails with "Invalid GeoIP database type, can't look up City by IP".
    I think this patch will be accepted upstream, because [GeoIP.country](https://github.com/cjheath/geoip/blob/c3f42995898f1ac92e948d5102d60e2888da25ca/lib/geoip.rb#L262) handles a superset of the database types supported by [GeoIP.city](https://github.com/cjheath/geoip/blob/c3f42995898f1ac92e948d5102d60e2888da25ca/lib/geoip.rb#L364).

I am not advocating that we do this; I still think geolocation is not worth the hassle. Just pointing out alternatives.

tom93:
Depends on #113 to fix build failure.

bagedevimo (dismissed):
Strongly in favour - this geocoder is a lot of complexity for a very small feature that the vast majority of users will never even see.

tom93:
Fixed merged conflicts.

bagedevimo (approved):

PR #103, opened 2020-04-17
Show all results publiclly.

puqeko:
Previously students results would only be shown on the public
scoreboard if equal to or better than the median score. I don't see any point to this - we may as well show the preliminary results publicly so people can keep an eye on the scoreboard without logging in.

PR #109, opened 2020-04-23
Update owner column in the contest index pages

tom93:
- Only display it to admins
 - Change the heading from "User" to "Owner"
 - Display the username (and link to their profile) instead of ID

Based on commits af301f9 (from #103) and 784e99c (from branch 'scoreboard-update').

I used `handle(contest.owner)` instead of `contest.owner.username` so it doesn't crash if the user has been deleted. A side effect is that it displays the full name of the user in addition to their username.

If the owner has been deleted it displays "DELETED". The link will point to a somewhat arbitrary page (e.g. on /contests it points to /contests/my, and on /contests/current it points to /contests/active; I'm not sure exactly why). This doesn't seem like a major issue, and it is already widespread (all the places that use `link_to handle(XXX), XXX` where `XXX` may be nil), so I'm ignoring it.

*Update: changed back to only displaying username, see https://github.com/NZOI/nztrain/pull/109#issuecomment-618441945*

### DIFF
--- a/app/views/contests/_index.html.erb
+++ b/app/views/contests/_index.html.erb
@@ -31,7 +31,7 @@
     <td><%= contest.end_time.strftime("%b %d, %H:%M") unless contest.end_time.nil? %></td>
     <td><%= contest.duration %></td>
     <% if policy(Contest).update? %>
-      <td><%= link_to handle(contest.owner), contest.owner %></td>
+      <td><%= link_to contest.owner.username, contest.owner if contest.owner.present? %></td>
     <% end %>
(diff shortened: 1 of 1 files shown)


Holmes98:
I'd prefer if it didn't show the full name, as it takes up quite a bit of extra space. The owner column in the problems index shows just the username, and just omits the owner if they've been deleted (see [here](https://github.com/NZOI/nztrain/blob/master/app/presenters/application_presenter.rb#L6)). Would you be fine with that instead?

Issue #106, opened 2020-04-22
Filelink visibility "protected" looks useless


tom93:
The visibility of filelinks can be set to "protected", which allows users to view the file if they are not currently competing (see [app/models/filelink.rb:17](https://github.com/NZOI/nztrain/blob/843d01e5a25965046fc06742beb615878065d11e/app/models/filelink.rb#L17), [app/policies/filelink_policy.rb:26](https://github.com/NZOI/nztrain/blob/843d01e5a25965046fc06742beb615878065d11e/app/policies/filelink_policy.rb#L26)).
I don't understand what this is for.

The problem series importer sets the visibility of problem statement attachments to "protected" (see [app/workers/problem_series/import_worker.rb:254](https://github.com/NZOI/nztrain/blob/843d01e5a25965046fc06742beb615878065d11e/app/workers/problem_series/import_worker.rb#L254)), but I think the intention there is for the statement to be visible only if the user *is* currently competing. (Fortunately, this doesn't expose the problem statements before the competition because the controller also checks that the containing problem is accessible, see [app/controllers/filelinks/roots_controller.rb:51](https://github.com/NZOI/nztrain/blob/843d01e5a25965046fc06742beb615878065d11e/app/controllers/filelinks/roots_controller.rb#L51:L63).)

I thought it might be to do with closed-book exams, but that doesn't make sense either because the policy doesn't check if the contest is closed book or open book.

Does anyone have any insights as to what "protected" is meant for?
If not, I propose that we remove "protected" (and change the problem series importer to use "public").

Holmes98:
> I thought it might be to do with closed-book exams, but that doesn't make sense either because the policy doesn't check if the contest is closed book or open book.

I think the intention *is* actually to prevent access to file attachments of other problems during closed-book contests (the functionality to choose which type doesn't seem to be implemented, so all contests are closed-book at the moment). But as you said, the problem authorization already handles this, and it doesn't work if a contest problem has protected file (since it disallows access to competing users), so it doesn't seem too useful.

The only case I can think of where it could potentially be useful is for file attachments in groups – in that case, `public` files would be visible, but `protected` files would not during a contest.

Whether we keep it or not (I'm undecided), I do agree with changing the importer to use `public`.

tom93:
Import worker fixed in #112.
Closing because it makes sense to use 'protected' for file attachments in groups to hide them during a contest.

PR #112, opened 2020-04-23
Change visibility of problem statement attachment to 'public' in importer

tom93:
From #106.

The problem series import worker previously set the visibility of the file attachment containing the imported statement to 'protected'.

That doesn't make sense, because 'protected' means users can view the file if they are *not* currently competing (comment: [app/models/filelink.rb:17](https://github.com/NZOI/nztrain/blob/843d01e5a25965046fc06742beb615878065d11e/app/models/filelink.rb#L17), logic: [app/policies/filelink_policy.rb:26](https://github.com/NZOI/nztrain/blob/843d01e5a25965046fc06742beb615878065d11e/app/policies/filelink_policy.rb#L26)).

As discussed in #106 (specifically https://github.com/NZOI/nztrain/issues/106#issuecomment-617556022 by @Holmes98), the purpose of 'protected' is probably to prevent access to file attachments of other problems/groups during closed-book contests. It does not make sense to use it for problem statements, because users need to be able to view them during the contest.

This commit changes the visibility to 'public'. It is safe (in the sense that it won't make the problem statements visible before the user starts the contest) because the controller also checks that the
containing problem is accessible (see [app/controllers/filelinks/roots_controller.rb:51](https://github.com/NZOI/nztrain/blob/843d01e5a25965046fc06742beb615878065d11e/app/controllers/filelinks/roots_controller.rb#L51:L63)).

### DIFF
--- a/app/workers/problem_series/import_worker.rb
+++ b/app/workers/problem_series/import_worker.rb
@@ -251,7 +251,7 @@ class ProblemSeries
         original_attachment = filelink.file_attachment
         filelink.file_attachment = nil
         filelink.file_attachment = FileAttachment.new(file_attachment: File.open(paths[:statement]), owner_id: 0, name: "coci-#{problem.name.parameterize}-statement.pdf")
-        filelink.visibility = Filelink::VISIBILITY[:protected]
+        filelink.visibility = Filelink::VISIBILITY[:public]
         if filelink.save
           log "Statement attached for problem #{data[:name]}."
         else


tom93:
To fix the visibility of existing file attachments, we can run this in the Rails console:
```ruby
Filelink.where(:visibility => Filelink::VISIBILITY[:protected]).each { |f|
  puts "filelink #{f.id} #{f.filepath}, attachment #{f.file_attachment_id} #{f.file_attachment.name}, root #{f.root_id} #{f.root_type} #{f.root.name}"
  # f.visibility = Filelink::VISIBILITY[:public]
  # f.save!
}
```
(The lines that modify the database are commented out. We should run it once as-is to see which filelinks have protected visibility. They should all be from COCI problems, if not we should investigate. Then we can uncomment the lines and run it again to actually change the visibility.)

tom93:
Deployed and updated database using console.
There were a few non-COCI problem statements with protected visibility, I changed them to public too.

PR #122, opened 2021-02-05
Fix redis-installer script

Holmes98:


tom93 (approved):
Approved.

This file is based on https://github.com/wayneeseguin/redis-installer, which is no longer maintained. We should probably switch to the [official Ubuntu PPA](https://redis.io/download#from-the-official-ubuntu-ppa) at some point to get automatic updates etc. but that can wait.

I had a quick look and I think the only part of this script we might need to keep is the random password generation, and even that might not be necessary. Redis only binds to localhost by default (see also: [protected mode](https://redis.io/topics/security#protected-mode)), so a password is only needed to protect from other users on the same machine. But our install script currently does not restrict access to the file /etc/redis/redis.conf (which contains the password), so the password isn't secret. (Note that on the server the permissions are tighter, presumably done manually.) We could restrict access to the file (e.g. make the file readable only by the "redis" group, and add the web app user to that group so it can [read the password](https://github.com/NZOI/nztrain/blob/6b0dc540b12e3fcfdae71c0f53dee0fa344f3474/config/initializers/redis.rb#L16)). Another approach might be to switch to a UNIX socket and use file permissions to restrict access to the socket instead of using a password.

PR #123, opened 2021-02-06
Update contest scoreboard

Holmes98:
Based on PR #103.

Notable changes:
- Unofficial contestants are unranked and hidden by default, and can be displayed by toggling a checkbox.
- If live scoreboard is unchecked, the entire scoreboard is now hidden (d4f97fd).
- Users can no longer set a school if their country is not New Zealand (aca4daa).
   - Add a migration to remove schools of users outside NZ (e4c35fe).


app/views/contests/scoreboard.html.erb
```diff
@@ -9,17 +9,27 @@
   //-->
   </script>
 <% end %>
+<% show_scoreboard = @contest.live_scoreboard || @contest.ended? || policy(@contest).inspect? %>
 <% if !@contest.started? %>
   <em>The contest has not started yet.</em>
+<% elsif !show_scoreboard && (!user_signed_in? || !@contest.has_competitor?(current_user)) %>
```
tom93, line 15: Would it be clearer if the negation was factored out?
```suggestion
<% elsif !(show_scoreboard || (user_signed_in? && @contest.has_competitor?(current_user))) %>
```
I think this makes it slightly clearer that the `else` part (which displays the scoreboard) will be executed if show_scoreboard is true or the user is competing.

app/views/contests/scoreboard.html.erb
```diff
@@ -29,20 +39,25 @@
...
-      <% median = @scoreboard[@scoreboard.length/2-1] %>
       <% rank = 1 %>
+      <% num_unofficial = 0 %>
       <% previous_record = @scoreboard.first %>
       <% @scoreboard.each_with_index do |record,index| %>
-        <% if !user_signed_in? or record.user && record.user.id != current_user.id && !policy(@contest).inspect? %>
-          <% next if record.score < median.score || (record.score == median.score && record.time_taken > median.time_taken) # no permission to view %>
-        <% end %>
-        <tr <% if user_signed_in? && record.user && record.user.id == current_user.id %> class="emphasized"<% end %>>
+        <% next if !show_scoreboard && (!user_signed_in? or record.user && record.user.id != current_user.id) # no permission to view %>
```
tom93, line 51: This inherits the bug fixed by #117 (deleted users will be shown to signed-in users when the scoreboard is not live). I'd also like to refactor as in #118, i.e.
```suggestion
        <% next unless show_scoreboard || (user_signed_in? && record.user && record.user.id == current_user.id) # hide row unless permitted to view %>
```
(Might be worth saving `user_signed_in? && record.user && record.user.id == current_user.id` in a variable, because it's also used below.)

app/views/contests/scoreboard.html.erb
```diff
@@ -29,20 +39,25 @@
...
+        <% is_unofficial = @contest.show_unofficial_contestants && (record.school.nil? || record.school_year.nil?) %>
+        <tr class="<% if user_signed_in? && record.user && record.user.id == current_user.id %>emphasized<% end %><% if is_unofficial %> unofficial<% end %>">
+          <% num_unofficial += 1 if is_unofficial %>
+          <% rank = index + 1 - num_unofficial unless previous_record.score == record.score && previous_record.time_taken == record.time_taken %>
```
tom93, line 55: I think there is a bug here. I haven't tested it, but I'm pretty sure the following scoreboard will trigger it:

| Rank | User           | Score |
|------|----------------|------:|
| 1    | A              | 100   |
| -    | B (unofficial) | 90    |
| 2    | C              | 90    |

I'm pretty sure the rank of user C will be (incorrectly) displayed as 1, because the rank isn't updated when the score is equal to the previous score.

How about:

```erb
# before loop
      <% num_ranked = 0 %>
      <% rank = nil %>
      <% previous_record = nil %>
# in loop
          <% if is_ranked %>
            <% num_ranked += 1 %>
            <% rank = num_ranked unless previous_record && previous_record.score == record.score && previous_record.time_taken == record.time_taken %>
            <% previous_record = record %>
          <% end %>
```

app/views/contests/scoreboard.html.erb
```diff
@@ -29,20 +39,25 @@
...
+          <% previous_record = record %>
+          <td style="text-align: right"> <%= rank unless is_unofficial || !show_scoreboard %> </td>
```
tom93, line 57: With the previous suggestion, this would become `rank if is_ranked && show_scoreboard`.

app/views/contests/scoreboard.html.erb
```diff
@@ -66,11 +81,23 @@
...
-          <td align="right"> <%= rank %> </td>
         </tr>
       <% end %>
     </tbody>
   </table>
 <% end %>
+<script>
+  function toggle_unofficial() {
+    show_unofficial_contestants = document.getElementById("show_unofficial").checked;
+    unofficial = document.getElementsByClassName("unofficial");
```
tom93, line 92: Missing `var`.

app/views/contests/scoreboard.html.erb
```diff
@@ -66,11 +81,23 @@
...
+    for (let row of unofficial) {
+      if (row.classList.contains("emphasized")) continue  // always show current user
```
tom93, line 94: Missing `;`

db/migrate/20210206104409_remove_schools_for_foreign_users.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveSchoolsForForeignUsers < ActiveRecord::Migration
+  def up
+    User.find_each do |user|
+      if user.country_code != "NZ" && user.school
+        user.school = nil
+        user.save
```
tom93, line 6: Shouldn't `user.save` be `user.save!` to fail loudly on errors?

tom93 (changes requested):
(Preliminary feedback.)

Is this PR intended to be squashed, or merged with the commit history? If we are keeping the history I'd like to clean it up (squash fixes, fix [spelling](https://github.com/NZOI/nztrain/pull/123/commits/bbb0a51c2109d5953a79a59afdb8f050dcfb282d) in commit messages, and insert my commits from #117 and #118).

app/views/contests/_form.html.erb
```diff
@@ -46,8 +53,16 @@
...
     <% end %>
-
   </div>
+  <div class="field">
+    <%= f.check_box :live_scoreboard %>
+    <%= f.label :live_scoreboard %> - If checked, other users' scores are shown during the contest. Otherwise, you may only see your own scores until the end of the contest.<br />
+  <div>
+  <div class="field">
+    <%= f.check_box :show_unofficial_contestants %>
+    <%= f.label :show_unofficial_contestants %> - If checked, unofficial contestants are not ranked. An offical contestant is any student enrolled in a NZ school.<br />
```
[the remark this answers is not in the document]
Holmes98: Sure, `only_rank_official_contestants` sounds good.

app/views/contests/scoreboard.html.erb
Holmes98: Good catch.

Holmes98:
@tom93 Thanks for the initial review, I'm fine with those changes.

I was intending to squash, since the history is still visible in this PR. Some commits (e.g. bbb0a51, aca4daa) can be moved into separate PRs, and I don't mind merging #117, #118 in their own PRs if you want to preserve their history in master.  Otherwise, if you prefer to merge, feel free to clean up the commit history of this PR and add your fixes.

tom93:
> I was intending to squash, since the history is still visible in this PR. Some commits (e.g. [bbb0a51](https://github.com/NZOI/nztrain/commit/bbb0a51c2109d5953a79a59afdb8f050dcfb282d), [aca4daa](https://github.com/NZOI/nztrain/commit/aca4daa07be86f88edfe0a68cf332c8a5cb35237)) can be moved into separate PRs, and I don't mind merging #117, #118 in their own PRs if you want to preserve their history in master. Otherwise, if you prefer to merge, feel free to clean up the commit history of this PR and add your fixes.

I'd like to keep some of the commits in the repo history. How about we keep working in this branch/PR, then create a new PR with the cleaned-up history and merge that. GitHub doesn't allow manually marking a PR as merged, so we'll have to can close the other PRs and add a comment saying they were merged as part of the cleaned-up PR.
I'll start implementing my suggestions and pushing to the branch "scoreboard-update-v2" (may include untested code).

app/views/contests/_form.html.erb
```diff
@@ -46,8 +53,16 @@
     <% else %>
       <%= handle(@contest.owner) %>
     <% end %>
-
   </div>
+  <div class="field">
+    <%= f.check_box :live_scoreboard %>
+    <%= f.label :live_scoreboard %>Show the scoreboard during the contest. If unchecked, users can only see their own score until the end of the contest.<br />
```
Holmes98, line 59: This results in the following text:

> Live scoreboardShow the scoreboard during the contest. If unchecked, users can only see their own score until the end of the contest.

Suggest either prepending " - " to the description as before, or just removing the "Live scoreboard" part:
```suggestion
    <%= f.label :live_scoreboard, "Show the scoreboard during the contest. If unchecked, users can only see their own score until the end of the contest." %><br />
```
(same with `only_rank_official_contestants`)

tom93: Oops, I didn't realise there was an automatic label too. I added the "-" back for now.
Replacing the automatic label ("Live scoreboard") with the long description as in your suggested edit makes the entire description clickable, which I don't like (too easy to accidentally toggle).
Maybe we should use parenthesised italics for the description, as done for the password field:
https://github.com/NZOI/nztrain/blob/84633ca11a3cac7d7d6fb9dffa885f8a90350b24/app/views/accounts/registrations/edit.html.erb#L15
I'm thinking of rewording `only_rank_official_contestants` so it displays as "Only rank official contestants - Exclude contestants from ranking if they are not enrolled in a NZ school."

Holmes98: I don't mind switching to parenthesised italics, maybe with some extra spacing between the label and description.

>  I'm thinking of rewording `only_rank_official_contestants` so it displays as "Only rank official contestants - Exclude contestants from ranking if they are not enrolled in a NZ school."

Sounds good.

eric-ide:
Perhaps the help text for only_rank_official_contestants should make it clear that viewers will still have the option to see unofficial contestants on the scoreboard? Currently it might give the impression that unofficial contestants will be removed from the scoreboard entirely.

app/views/contests/info.html.erb
```diff
@@ -1,20 +0,0 @@
...
-  <h2>Contest Registration</h2>
-
-  <strong>Country: </strong><%= @contest_relation.country_code %>
-  <strong>School: </strong><%= @contest_relation.school.name %>
-<% end %>
-
-<h2>Contest Settings</h2>
-<b>Mode: </b>Closed book (you cannot access other problems on the site while competing)<br>
-<b>Judge feedback: </b>Real-time<br>
-<b>Scoreboard: </b>Live<br>
```
tom93, line 19: I agree that the "info" tab was pretty useless before, but I don't think we should delete it. (I'm fine with redirecting to the "scoreboard" tab by default instead of "info" for non-signed-in users though.)

 - The line "Mode: Closed book" seems useless because the site doesn't support open-book contests, but I think we should keep it so first-time users know they won't have access to their previous submissions. I'm OK with removing this line though.
 - The line "Judging feedback: Real-time" can be removed (judging feedback is always real-time on our site, and also almost everywhere else these days -- the only exception I can think of is Google Code Jam's hidden verdict test cases).
 - "Scoreboard: Live" was previously always the case, but we are adding non-live scoreboards so this line becomes important. Without it, users can't tell whether past contests had live scoreboards or not. Also, it might help users during the contest, if they wonder why there is nobody else on the scoreboard (though they should have already seen the message "The scoreboard is hidden until the end of the contest" before starting). I think we should also add an explanation of hidden scoreboards here, e.g. "Scoreboard: Hidden (you can only see your own score until the end of the contest)".
 - We might want to move some of the information from the rest of the page into the "info" tab, to take up less space on the other tabs. I'm think specifically of the "Owner" line and the finalize/un-finalize buttons, but we might also want to move the start time, end time, and duration (at first I thought they were too important, but then I realised the user probably already saw that information in the contest list).

app/views/contests/_admin.html.erb
```diff
@@ -16,3 +16,7 @@
 <% end %>
 </ul>
 
+<p>
+  <b> Problem set: </b>
+  <%= link_to @contest.problem_set.name, @contest.problem_set %>
```
tom93, line 21: How about moving the problem set link to the top of the "problems" tab? As written it takes up space on all tabs, including the scoreboard.

(Maybe in separate PR we could also move the existing groups UI into one of the tabs, I'm thinking either the "contestants" tab, or a new "groups" tab; if we use the "contestants" tab, my preferred UI is: list of groups with access, button to add a group, button to register a single user, then list of contestants)

app/views/contests/scoreboard.html.erb
```diff
@@ -66,11 +84,23 @@
...
       <% end %>
     </tbody>
   </table>
 <% end %>
+<script>
+  function toggle_unofficial() {
+    var show_unofficial_contestants = document.getElementById("show_unofficial").checked;
+    var unofficial = document.getElementsByClassName("unofficial");
+    for (let row of unofficial) {
+      if (row.classList.contains("emphasized")) continue;  // always show current user
```
tom93, line 97: We should use a more meaningful class name, e.g. "current-user" (in addition to the "emphasized" class, which is for styling).

app/views/contests/_form.html.erb
```diff
@@ -46,8 +53,16 @@
...
+    <%= f.label :live_scoreboard %> - Show the scoreboard during the contest. If unchecked, users can only see their own score until the end of the contest.
+  </div>
+  <div class="field">
+    <%= f.check_box :only_rank_official_contestants %>
+    <%= f.label :only_rank_official_contestants %> - Exclude unofficial contestants from ranking. An official contestant is any student enrolled in a NZ school.
+  </div>
+  <br>
   <div class="actions">
     <%= f.submit %>
   </div>
```
tom93, line 68: This `<br>` adds space between the last checkbox and the "Create Contest" button, is it intentional? There wasn't added space before, and I don't think the added space is necessary/consistent.

tom93 (changes requested):
> Perhaps the help text for only_rank_official_contestants should make it clear that viewers will still have the option to see unofficial contestants on the scoreboard? Currently it might give the impression that unofficial contestants will be removed from the scoreboard entirely.

Good point. How about appending "Users will have an option to show or hide unofficial contestants when they view the scoreboard."
Or just changing "ranking" to "ranking calculation", so the message would be "Only rank official contestants - Exclude contestants from ranking calculation if they are not enrolled in a NZ school."

app/views/contests/info.html.erb
Holmes98: I don't mind restoring the info tab. The owner field can be moved. I feel like finalize should appear on the scoreboard tab (since it basically locks the state of the scoreboard), so I'd prefer to leave it or move it to appear only on the scoreboard tab. I'd would prefer to keep the others visible on all tabs (users could forget or miss it, and they might have been linked directly to a contest page).

app/views/contests/_admin.html.erb
Holmes98: I'm fine with that.

app/views/contests/scoreboard.html.erb
Holmes98: Sounds good.

app/views/contests/_form.html.erb
Holmes98: I agree, it can be removed.

Holmes98:
>  Or just changing "ranking" to "ranking calculation", so the message would be "Only rank official contestants - Exclude contestants from ranking calculation if they are not enrolled in a NZ school."

This sounds good.

Holmes98:
Also note that I'm planning to merge the validation part of aca4daa in #129 instead. I'll likely remove the school migration and just do it manually since I also want to clean up the schools list (removing non-NZ schools).

app/views/contests/info.html.erb
tom93: I restored the info tab and added the explanation of hidden scoreboards.

I suggest that we separately remove the line "Judging Feedback: Real-time", move the "Owner" line into the info tab, and move the finalize/un-finalize buttons into the scoreboard tab.

app/views/contests/_admin.html.erb
tom93: Done. While doing this, I noticed invalid `<p>` tags around the problems table, so I removed them (see commit df5c0536f4affe0af6830940c522c30fc1497f99). They had the effect of adding blank space above and below the problems table. I don't think it's intentional (the scoreboard and supervisor tables don't such space). Here is what it looks like before and after.

Before (with space above and below the problems table due to `<p>` tags):

1-old

After removing the `<p>` tags:

2-new

After also moving the problem set link:

3-new-moved-problem-set

Holmes98: Looks good to me.

app/views/contests/info.html.erb
Holmes98: Sure, I'm fine with that.

Holmes98:
Will deploy the current state of this PR in branch `deployed` (7436797) once NZIC R1 has ended.

tom93 (approved):
Approved.

I've only noticed just now that unofficial contestants are very easy to miss because they are hidden by default and the checkbox to toggle their visibility is far on the right side of the screen. I'm not sure if this is desirable or not.

An idea to increase their visibility is to move the checkbox to the left side of the screen, next to the scoreboard heading:

scoreboard-checkbox-float-left

(done by moving the `<div>` element with the checkbox to be after the `<h2>` element, then floating both the `<h2>` and the `<div>` to the left and adding a margin between them)

That does make the heading area a bit too "busy" though. Leaving it up to you to decide.

tom93:
For future reference, Rails database migrations can be [generated](https://guides.rubyonrails.org/v3.2/migrations.html#creating-a-standalone-migration), and the generator uses a specific naming convention ("AddColumnToTable"). For example

    rails generate migration AddLiveScoreboardToContests live_scoreboard:boolean

creates a file called `db/migrate/20210228174106_add_live_scoreboard_to_contests.rb` with the following contents:

```ruby
class AddLiveScoreboardToContests < ActiveRecord::Migration
  def change
    add_column :contests, :live_scoreboard, :boolean
  end
end
```

This is just missing `:default => true`, which [needs to be added manually](https://stackoverflow.com/questions/24565589/can-i-pass-default-value-to-rails-generate-migration).

PR #125, opened 2021-02-07
Add interfaces for managing schools

Holmes98:
Currently any user can add a school to the school list, so this is required to allow for moderation. I did some cleanup last year through `rails console` but that isn't really ideal.
In the future, we might want to consider restricting additions to the list.

### DIFF
--- a/app/controllers/contests_controller.rb
+++ b/app/controllers/contests_controller.rb
@@ -307,7 +307,9 @@ class ContestsController < ApplicationController
               redirect_to contestants_contest_path(@contest), :alert => "The maximum extra time that can be given is #{@contest.max_extra_time}."
               return
             end
-            relation.save
+            if !relation.save
+              redirect_to contestants_contest_path(@contest), :alert => "Could not add extra time to contestant #{relation.user&.username}."
+            end
           end
         end
       end
--- a/app/controllers/contests_controller.rb
+++ b/app/controllers/contests_controller.rb
@@ -293,6 +293,7 @@ class ContestsController < ApplicationController
             relation.school = School.find_by_id(school_id)
             if !relation.save
               redirect_to contestants_contest_path(@contest), :alert => "Could not update school of contestant #{relation.user&.username}."
+              return
             end
           end
         end
@@ -309,6 +310,7 @@ class ContestsController < ApplicationController
             end
             if !relation.save
               redirect_to contestants_contest_path(@contest), :alert => "Could not add extra time to contestant #{relation.user&.username}."
+              return
             end
           end
         end
--- a/app/controllers/contests_controller.rb
+++ b/app/controllers/contests_controller.rb
@@ -285,6 +285,18 @@ class ContestsController < ApplicationController
       end
       redirect_to contestants_contest_path(@contest), :notice => "Contest ended for selected students"
     elsif params[:update]
+      if params[:school_id]
+        ContestRelation.transaction do
+          params[:school_id].each do |relation_id, school_id|
+            relation = @contest.contest_relations.find_by_id(relation_id)
+            authorize relation, :update_school?
+            relation.school = School.find_by_id(school_id)
+            if !relation.save
+              redirect_to contestants_contest_path(@contest), :alert => "Could not update school of contestant #{relation.user&.username}."
+            end
+          end
+        end
+      end
       if params[:extra_time]
         ContestRelation.transaction do
           params[:extra_time].each do |relation_id, extra_time|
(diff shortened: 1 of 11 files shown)


app/policies/contest_relation_policy.rb
```diff
@@ -28,5 +28,9 @@ def update_extra_time?
     return false if record.contest.ended?
     supervise?
   end
+
+  def update_school?
+    user.is_admin?
```
puqeko, line 33: Does this give permissions to super admins as well or just admins specifically?

app/controllers/contests_controller.rb
```diff
@@ -285,6 +285,18 @@ def supervise
       end
       redirect_to contestants_contest_path(@contest), :notice => "Contest ended for selected students"
     elsif params[:update]
+      if params[:school_id]
```
puqeko, line 288: Cool, but do we need this feature? If we select their username from the contestant list, then click edit, we can edit their school that way.

app/controllers/contests_controller.rb
```diff
@@ -298,7 +310,9 @@ def supervise
             relation.save
```
puqeko, line 310: Should we add a similar check for `if !relation.save` in the `:extra_time case` to be consistent with your addition above?

app/views/schools/index.html.erb
```diff
@@ -1,15 +1,31 @@
...
+    </tr>
   </thead>
   <tbody>
     <% @schools.each do |school| %>
       <tr>
         <td><%= school.id %></td>
         <td><%= link_to school.name, school %></td>
         <td><%= school.users_count %></td>
+        <% if policy(School).update? %>
+          <td><%= link_to 'Edit', edit_school_path(school) if policy(school).update? %></td>
```
puqeko, line 24: Note that on the to school edit page the <- back button sends a person to the specific school page with the user list rather than back to the schools page the person came from. The path takes two back buttons to get to where a person starts: [schools list page] -> (edit) -> [school edit page] -> (back) -> [school page] -> (back) -> [schools list page]. I personally don't mind leaving this behaviour if it is too hard to fix though

puqeko (changes requested):

puqeko:
How hard would it be to add a "New" button to the schools list page to be consistent with other pages? Having to create new users to add schools is a bit awkward (unless I've missed something).

Also, could we update the registration page to grey out the "School" part if they do not pick NZ as their country? It would also be nice if "School Graduation" stated that the unspecified option is not eligible for both IOI and NZIC.

Screen Shot 2021-02-11 at 7 33 22 PM

app/controllers/contests_controller.rb
Holmes98: It's slightly different, this is editing the school of the contest relation, whereas the other one is editing the school of the user. They're stored separately because a user might change schools after participating in a contest, and we want to preserve whichever school they were in at the time of the contest.

app/views/schools/index.html.erb
Holmes98: Good point, it should use `:back` instead (430b878)

app/policies/contest_relation_policy.rb
Holmes98: It checks for both `:admin` and `:superadmin`
https://github.com/NZOI/nztrain/blob/84633ca11a3cac7d7d6fb9dffa885f8a90350b24/app/models/user.rb#L120-L121

Holmes98:
> How hard would it be to add a "New" button to the schools list page to be consistent with other pages? Having to create new users to add schools is a bit awkward (unless I've missed something).

Not sure, but you can also add a school by editing your profile (though you'll then need to change it back afterwards). You typically don't need to add schools though, since users can add them on registration. You'll mostly only need to delete them (because someone added a fake/duplicate school), or rename them.

> Also, could we update the registration page to grey out the "School" part if they do not pick NZ as their country?

It would be nice, though I'm pretty sure this requires JavaScript. I won't add it myself, but I don't mind if anyone else wants to.

> It would also be nice if "School Graduation" stated that the unspecified option is not eligible for both IOI and NZIC.

Sure. It should probably also clarify that it's the expected high school graduation date (people often ask "What if I haven't graduated yet?", and some are in primary/intermediate school).

app/controllers/contests_controller.rb
Holmes98: I don't mind, though I don't think there are any cases where validation would fail, I just added the check to be extra safe.

app/controllers/contests_controller.rb
puqeko: Ah, okay. Good to know

app/policies/contest_relation_policy.rb
puqeko: Sweet

puqeko:
> It should probably also clarify that it's the expected high school graduation date (people often ask "What if I haven't graduated yet?", and some are in primary/intermediate school).

Yes, that's a good point.

app/views/schools/index.html.erb
puqeko: The school edit page already uses `:back`. Not sure what's happening there

Holmes98: It's the second parameter that needs to be `:back`. Resolved in 946ebaa.

puqeko (dismissed):

puqeko (approved):

PR #128, opened 2021-02-12
Update registration page

puqeko:
* Expand detail on graduation date and eligibility
* Reformat to use headers instead of paragraphs
* <strike>Make "new school" slightly more noticeable</strike>
* Improve spacing of header for small devices

Avoided changing any stylesheets for fear of impacting other pages

<strike>Merge after PR #125 </strike>

### DIFF
--- a/app/views/accounts/registrations/new.html.erb
+++ b/app/views/accounts/registrations/new.html.erb
@@ -1,45 +1,63 @@
-<h2>Sign up</h2>
-
+<h1 style="line-height:1.2em">Sign up to the NZOI's training site</h1>
+<ul>
+  <li>Submit solutions to problems.</li>
+  <li>Compete in contests.</li>
+  <li>Join groups.</li>
+  <li>Access more problems.</li>
+</ul>
 <%= form_for(resource, :as => resource_name, :url => registration_path(resource_name)) do |f| %>
   <%= devise_error_messages! %>
 
-  <p><%= f.label :username %><br />
+  <h3><%= f.label :username %><br />
   <%= f.text_field :username %></p>
 
-  <p><%= f.label :name %><br />
+  <h3><%= f.label :name %><br />
   <%= f.text_field :name %></p>
 
-  <p><%= f.label :email %><br />
+  <h3><%= f.label :email %><br />
   <%= f.email_field :email %></p>
 
-  <p><%= f.label :password %><br />
+  <h3><%= f.label :password %><br />
   <%= f.password_field :password %></p>
 
-  <p><%= f.label :password_confirmation %><br />
+  <h3><%= f.label :password_confirmation %><br />
   <%= f.password_field :password_confirmation %></p>
 
-  <p><%= f.label :country_code, "Country" %><br />
+  <h3><%= f.label :country_code, "Country" %><br />
   <%= f.country_select(:country_code, selected: @user.country_code || request.location.try(:country_code), priority_countries: ["NZ", "AU"]) %></p>
 
-  <p><%= f.label :school_graduation, "School Graduation" %><br />
+  <h3><%= f.label :school_graduation, "School Graduation" %></h3>
+  <p> If you attend a primary, intermediate, or secondary school (or equivilent) in New Zealand then we ask that you please enter your estimated graduation date (the month and year you expect to finish secondary school by). You can update this later if it changes. Our website uses your graduation date to calculate your eligibility to:ssss</p>
+  <ul>
+    <li>Compete <u>officially</u> in the NZIC (New Zealand Informatics Contest).</li>
+    <li>Potentally be invited to a programming camp run by the NZOI (New Zealand Olympiad in Informatics)</li>
+  </ul>
+  <p>Also note that:</p>
+  <ul>
+    <li>Anyone is welcome to create an account and use our site without specifying a graduation date.</li>
+    <li>Anyone can compete in the NZIC unofficially without specifying a graduation date. However, you will not be ranked on offical scoreboards or be eligible to be invited to our programming camps.</li>
+    <li>Only New Zealand citizens or perminant residents, with a valid future graduation date, will be considered for our international IOI team.</li>
+  </ul>
+  <p><b>Please select approximate secondary school graduation date or leave unspecified:</b></p>
   <%= f.fields_for :school_graduation do |graduation_form| %>
-  <%= graduation_form.radio_button('enabled', false, checked: !@user.school_graduation) %><%= f.label(:school_graduation_enabled_false, "Unspecified (Not considered eligible for IOI)") %> <br />
+  <%= graduation_form.radio_button('enabled', false, checked: !@user.school_graduation) %><%= f.label(:school_graduation_enabled_false, "Unspecified") %> <br />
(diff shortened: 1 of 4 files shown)


puqeko:
Screen Shot 2021-02-12 at 9 05 25 PM

app/views/accounts/registrations/new.html.erb
```diff
@@ -1,45 +1,63 @@
...
     <a class="school_select togglelink jsnodisplay" data-toggle="school_select" href="javascript:void(0)">Select school from list</a>
-    <a class="school_select togglelink jsdisplay" data-toggle="school_select" href="javascript:void(0)">Add new school</a>
+    <a class="school_select togglelink jsdisplay" data-toggle="school_select" style="color:blue" href="javascript:void(0)">Add new school</a>
   <% end %>
   </p>
 
   <p>
     <%= recaptcha_tags :site_key => @db_settings["recaptcha/public_key"] if @db_settings["recaptcha/public_key"] && !@db_settings["recaptcha/public_key"].empty? %>
   </p>
+  <h3>Submit</h3>
```
Holmes98, line 60: Move the reCAPTCHA below "Submit" instead of "School".
```suggestion
  <h3>Submit</h3>
  <p>
    <%= recaptcha_tags :site_key => @db_settings["recaptcha/public_key"] if @db_settings["recaptcha/public_key"] && !@db_settings["recaptcha/public_key"].empty? %>
  </p>
```

app/views/accounts/registrations/new.html.erb
```diff
@@ -1,45 +1,63 @@
...
+  <p> If you attend a primary, intermediate, or secondary school (or equivilent) in New Zealand then we ask that you please enter your estimated graduation date (the month and year you expect to finish secondary school by). You can update this later if it changes. Our website uses your graduation date to calculate your eligibility to:ssss</p>
+  <ul>
+    <li>Compete <u>officially</u> in the NZIC (New Zealand Informatics Contest).</li>
+    <li>Potentally be invited to a programming camp run by the NZOI (New Zealand Olympiad in Informatics)</li>
+  </ul>
+  <p>Also note that:</p>
+  <ul>
+    <li>Anyone is welcome to create an account and use our site without specifying a graduation date.</li>
+    <li>Anyone can compete in the NZIC unofficially without specifying a graduation date. However, you will not be ranked on offical scoreboards or be eligible to be invited to our programming camps.</li>
+    <li>Only New Zealand citizens or perminant residents, with a valid future graduation date, will be considered for our international IOI team.</li>
```
Holmes98, line 39:
```suggestion
    <li>Anyone can compete in the NZIC unofficially without specifying a graduation date. However, you will not be ranked on official scoreboards or be eligible to be invited to our programming camps.</li>
    <li>Only New Zealand citizens or permanent residents that have attended camp will be considered to represent New Zealand at the International Olympiad in Informatics.</li>
```

Holmes98 (changes requested):
Do we also want to change the "Edit Profile" page? (`app/views/user/_form.html.erb`)

app/views/accounts/registrations/new.html.erb
```diff
@@ -1,45 +1,63 @@
...
 
-  <p><%= f.label :school %><br />
+  <h3><%= f.label :school %></h3>
   <%= f.fields_for :school, @user.school do |school_form| %>
     <%= school_form.text_field :name, class: [:school_select, :jsnodisplay] %>
     <%= f.select :school_id, School.where(country_code: "NZ").order(:name).collect {|s| [s.name, s.id]}, { include_blank: true }, class: [:school_select, :jsdisplay], disabled: true %>, 
     <%= school_form.country_select(:country_code, only: ["NZ"] ) %><br />
     <a class="school_select togglelink jsnodisplay" data-toggle="school_select" href="javascript:void(0)">Select school from list</a>
-    <a class="school_select togglelink jsdisplay" data-toggle="school_select" href="javascript:void(0)">Add new school</a>
+    <a class="school_select togglelink jsdisplay" data-toggle="school_select" style="color:blue" href="javascript:void(0)">Add new school</a>
```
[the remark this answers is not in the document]
puqeko: Okay. The black hover is indeed disgusting. I have never liked the grey links though

app/views/accounts/registrations/new.html.erb
puqeko: Good spot

app/views/accounts/registrations/new.html.erb
puqeko: Hmmm. I need to start typing things in something that does spell checking

app/views/accounts/registrations/new.html.erb
```diff
@@ -1,42 +1,61 @@
...
   <%= graduation_form.radio_button('enabled', true, checked: !!@user.school_graduation) %><%= select_month(@user.school_graduation || 12, {prefix: 'user[school_graduation]'}, class: :radio_input) %><%= select_year(@user.school_graduation || Date.today, {start_year: [Date.today.year-50,@user.school_graduation.try(:year) || Date.today.year].min, end_year: Date.today.year+18, prefix: 'user[school_graduation]'}, class: :radio_input) %>
-  <% end %></p>
+  <% end %>
 
-  <p><%= f.label :school %><br />
+  <h3><%= f.label :school, "School (NZ Only)" %></h3>
   <%= f.fields_for :school, @user.school do |school_form| %>
     <%= school_form.text_field :name, class: [:school_select, :jsnodisplay] %>
-    <%= f.select :school_id, School.where(country_code: "NZ").order(:name).collect {|s| [s.name, s.id]}, { include_blank: true }, class: [:school_select, :jsdisplay], disabled: true %>, 
-    <%= school_form.country_select(:country_code, only: ["NZ"] ) %><br />
```
Holmes98, line 34: Removing this causes new schools to be added with a nil `country_code`, so they won't appear under forms which use `School.where(country_code: "NZ")`.

Holmes98: Should be fixed but someone else should probably review this.

puqeko: Hmmm. Good spot. A terrible way to solve this could be to add it back in but make it a hidden element on the page.

puqeko: If all schools are now only NZ schools then perhaps it is worth finding and replacing all instances of `School.where(country_code: "NZ")` and creating a db migration to drop this information. If so, I'd prefer to make this a separate pull request and reintroduce the field here so that this one doesn't get held up

Holmes98: Sorry for not being clearer, but I already added a fix in c07c9f6. I just thought someone else should review my changes before we merge. c1378fa can be reverted.

puqeko: Oh, okay, sorry. In that case, do we also need to change the `school` method in the `User` class too?

puqeko: Clarification from @Holmes98 chat:

> If we remove the country code select box, POST requests will no longer contain the `user[school][country_code]` parameter (so it's treated as `nil`). Previously, this was used in the `school=` method to query for a school with that country code (the `first_or_initialize` part adds a new school if none was found). The fix is to just always use 'NZ' instead of that parameter.

Holmes98 (approved):

PR #132, opened 2021-02-16
Add missing space and remove repeat label

puqeko:
Not sure how I missed this before sorry but I think it would look better with these simplifications


puqeko:
Assets were not recompiled on restart. Fixed now

PR #135, opened 2021-02-20
Disable async loading of JavaScript in production mode

Holmes98:
The async attribute seems to delay execution of scripts, even if they are cached.

This can cause some page elements to flash on load:
 - The left sidebar (if minimised)
 - Dropdown menu arrows
 - The school fields, if the user's country is not set to NZ (in #129)
 
In production mode, the application scripts are concatenated into a single file (currently about 290 KB), and it will typically be cached, so I don't think async is necessary.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -3,7 +3,7 @@
   <head>
       <link rel="shortcut icon" href="/favicon.ico">
       <%= stylesheet_link_tag "application" %>
-      <%= javascript_include_tag "application", :async => Rails.env.production? %>
+      <%= javascript_include_tag "application", :async => false %>
       <%= javascript_include_tag "//cdnjs.cloudflare.com/ajax/libs/mathjax/2.1.0/MathJax.js?config=TeX-AMS_HTML", :async => true %>
       <%= stylesheet_link_tag "//fonts.googleapis.com/css?family=Ubuntu:400italic,Ubuntu:400"%>
       <noscript><style type="text/css">.js_only { display: none; }</style></noscript>


Holmes98:
Deployed for testing in branch `deployed` (85136d1).

Removing async seems to solve the flashing issue on most pages, but it still occurs on larger pages such as the contest scoreboards or the user/problem index pages.

tom93 (approved):
Approved, even though this is not ideal, because I think it's better than the previous behaviour and further improvements can wait.

Loading the scripts synchronously is not ideal because it delays page load. I agree that it is better than the previous behaviour (async script), which had an early first paint followed by a later redraw (i.e. flash) when the script finished loading.

I've written some notes below about ways to improve the page load speed and I hope to implement them some day. They are quite long; the low-hanging fruit are the first section (eliminating the flashes entirely when using synchronous scripts) and the last section (Firefox's caching behaviour and refreshes).

---

**Eliminating the flashes entirely** (when using synchronous scripts)

The reason this PR doesn't completely eliminate the flashing in large pages is that the code only runs once all the DOM has been loaded (`$(document).ready()`), and the browser can start to render the page before that happens.

To completely eliminate the flashing, the code needs to run earlier &ndash; right after the relevant elements are parsed and loaded. This can be done by putting the code in a small inline `<script>` element after the last relevant start tag. In the case of the sidebar for example this is in app/views/layouts/scaffold.html.erb after the tag `<div id="main" ...>`.

If the main script is loaded synchronously (as in this PR), we can simply move the relevant code from the `$(document).ready()` call in app/assets/javascripts/application.js to small inline `<script>` elements in the HTML. I tested this technique on the sidebar and there were no flashes (I used Firefox on a local server; haven't tried other browsers).

If the main script is loaded asynchronously it's more complicated, mainly because jQuery might not be loaded yet; see the following section for a solution.

---

**Fast page load with \<script defer\> and a small synchronous script**

It would be nice if the main script did not block rendering, to improve the page load speed. This applies in particular when the main script is being fetched for the first time, but I think it will also make a difference when the script is already in the cache because it still needs to be executed. (See also the section "Firefox's caching behaviour and refreshes" below.)

The proper solution is to use a `defer` script for the bulk of the code, and a separate synchronous script for the code that modifies the page. Ideally, the synchronous script would be very small and not depend on jQuery.

In the case of the sidebar, it is pretty easy to minimise on load using a small self-contained JavaScript snippet (it would be even easier if we switched from jQuery animations to CSS transitions, see the section below). Here is an example implementation:

```js
function setupSidebar() {
  if (/(^|; *)expanded=false($|;)/.test(document.cookie)) {
    document.getElementById("side").style.width = '50px';
    document.getElementById('main-menu').style.left = '50px';
    document.getElementById('side-minify').innerHTML = "&#9654;";
    document.getElementById('main').style.left = '50px';
    document.getElementById('side-expanded').style.opacity = '0';
    document.getElementById('side-contracted').style.opacity = '1';
  }
}
```

This is basically a translation of the `shrink()` function to use vanilla JavaScript instead of jQuery (without animations, because this is for page load). It also uses dirty regex cookie parsing instead of the `readCookie()` function. The function would then be called from an inline script element in scaffold.html.erb after the tag `<div id="main" ...>`. To keep the code reasonably organised, the definition of the function can be placed near the other script elements in application.html.erb (either in an inline script element or an external synchronous script).

I'm not sure how difficult it will be to do the same for the dropdown menus and the school fields.

Another approach is to dynamically generate the HTML on the server side; see @eric-ide's branch "server-sidebar-init" (commit 916d670a011a). I think his code is incomplete, but the idea of checking the cookie on the server side is useful. Switching from jQuery animations to CSS transitions would make his approach easier. The two approaches are closely related in my view: they both solve the problem of running code early during page load. Tasks such as checking a cookie or adding a CSS class to an element can be done equally well using either approach (i.e. on the client or on the server).

The reason for `defer` for the bulk of the code rather than `async` is that `defer` scripts run later (see the diagram in the [spec](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async)). The bulk of the code should run as late as possible, because it does not affect the rendered page and running it early would slow down the other tasks. Note that `defer` scripts are executed before the DOMContentLoaded event is fired, so the synchronous script must not wait for the DOMContentLoaded event; it should run after the relevant elements have been parsed (as in the previous section).

---

**The curious case of \<script async\> in Chrome**

The previous section describes the best solution in my opinion, but if splitting up the code is too much effort we could return to a single async script (as before this PR) and add a simple workaround improve the behaviour in Chrome.

The async attribute is supposed to cause a script to "be fetched in parallel to parsing and evaluated as soon as it is available" ([spec](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async), [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-async)). A reasonable expectation is that if the script is cached, it would be evaluated immediately (like synchronous scripts) since there is nothing to fetch. The difference should only be when the script is not in the cache &ndash; in that case the async attribute allows the HTML parser to continue while the script is being fetched, leading to an earlier first paint. This is pretty good.

Unfortunately, as Holmes98 observed, Chrome doesn't behave this way. For some reason, Chrome assigns low loading priority to async scripts (see <https://addyosmani.com/blog/script-priorities/>; the official live documentation appears to be the Google Doc at [bit.ly/script-scheduling](https://docs.google.com/document/d/16rHWLu-0abC9WWLhLBFlIRtbSnOFzhKAXsCamsp0oAs/preview)\*). There was a [proposal to increase the priority](https://bugs.chromium.org/p/chromium/issues/detail?id=690550), but it was dropped in favour of [Priority Hints API](https://bugs.chromium.org/p/chromium/issues/detail?id=821464) (which is still a draft).

Fortunately, there is a workaround/hack to assign a high loading priority to async scripts in Chrome: use `<link rel=preload>` in addition to `<script async>`. This should reduce the flashing (compared to a plain async script), and should lead to an earlier first paint compared to the synchronous script in this PR.

There may still be some flashing, and [this comment](https://bugs.chromium.org/p/chromium/issues/detail?id=690550#c1) explains that executing an async script early can delay the first paint. The way to fix these problems is to use the solution described in the previous section (split the code into a small synchronous script that modifies the page and a large `defer` script for the rest of the code). But if that solution is too much effort, then the `<link rel=preload>` workaround should provide pretty good results.

<sup>\*I made the link to the Google Doc start in "preview" mode. To get the full Google Docs interface, change "preview" to "edit" in the URL (the doc has lots of suggested edits that make it super messy).</sup>

---

**CSS transitions instead of jQuery animations for sidebar**

The animated sidebar minimisation is currently implemented using jQuery. Browsers have had good support for CSS transitions for a long time, so we could safely switch to CSS. The sidebar would then be expanded/collapsed from JavaScript (or on the server side) by toggling a CSS class.

(It will probably be necessary to conditionally enable the transitions, to prevent animations while the page is loading. This should be simple to implement, e.g. enable transitions based on an "animate" class and only add that class when the minimise button is clicked for the first time.)

---

**Firefox's caching behaviour and refreshes**

In Firefox, when the user navigates to a page and a subresource (e.g. a script) is cached, the browser does not make a network request for that subresource. This is good. However, when the user refreshes the page (using Ctrl+R or the refresh button), the browser will send a network request for the subresource even if it is cached. The server will respond with HTTP 304 Not Modified (and won't send the entire file again), but this still takes up a full network round trip. So making the script synchronous (this PR) delays page load by a network round trip when a Firefox refreshes the page. It's not a big deal, but it's undesirable.

The solution is to use the HTTP header [Cache-Control: immutable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#immutable).

(Chrome does not respect this header, and instead uses heuristics to skip network requests on refresh. See <https://stackoverflow.com/a/41958007> and <https://bugs.chromium.org/p/chromium/issues/detail?id=611416#c57>.)

This header should be fine for the files in the /assets subtree, because they either include a content hash in the filename or the URL includes a version in the query string, so the content of any URL in the /assets subtree should never change.

The way to add this header is using the nginx configuration. The template in Git (script/install/nginx.app.conf) already has a section for caching /assets and has a line

```
add_header Cache-Control public;
```

so I think we just need to change this line to\*

```
add_header Cache-Control "public, immutable";
```

The configuration file on the server (in /etc/nginx/sites-available) does not have a section for caching /assets, so I guess we'll need to carefully add that section.

<sup>\*I think the quotes are necessary because of the space (even if not, they shouldn't hurt). I can't find any official documentation on the quoting rules for nginx configuration files; the closest I can find is https://serverfault.com/questions/793550/when-do-you-have-to-use-quotes-in-nginx-configuration.</sup>

PR #142, opened 2022-01-15
Add contest exporter

Holmes98:


app/exporters/contests/contest_exporter.rb
```diff
@@ -0,0 +1,100 @@
...
+
+        contest.scoreboard.each do |record|
+          is_ranked = !contest.only_rank_official_contestants || (record.school && record.school_year)
+          next unless is_ranked || include_unranked_contestants
+
+          if is_ranked
+            num_ranked += 1
+            rank = num_ranked unless previous_record && previous_record.score == record.score && previous_record.time_taken == record.time_taken
+            previous_record = record
+          end
```
tom93, line 26: This complex logic is the same as in app/views/contests/scoreboard.html.erb, it should really be refactored.
But it can wait for separate PR in the interest of moving forward with other things.

app/views/contests/_admin.html.erb
```diff
@@ -16,3 +16,11 @@
...
 </ul>
 
+<% if policy(Contest).export? %>
+  <br>
+  <b> Export Contest</b>
+  <%= form_tag( export_contest_path(@contest), :method => :get) do %>
+    <%= submit_tag "Download" %>
+  <% end %>
+<% end %>
+
```
tom93, line 26: I'd prefer for the export UI and the existing admin UI to move into "info" tab (https://github.com/NZOI/nztrain/pull/123#discussion_r578550681), I'll make a separate PR.

tom93 (changes requested):
I pushed a couple of proposed changes (completely untested), let me kow what you think:
 - add-contest-exporter-v3a-small-fixes-wip @ c11d30e [multiple commits]
 - add-contest-exporter-v3b-refactor-get-submissions-wip @ d51c5df
 - add-contest-exporter-v3c-efficient-submissions-export-wip @ 88786ea

What are the use cases for the export feature? The format is very rough:
 - The JSON files are a raw database dump using the internal representation, which is not user-friendly and seems inconvenient when using from a script. Would it be worthwhile to design a better format?
 - The submissions are currently in a single flat directory (with the submission ID as the filename), which is not user-friendly. It would be much easier to browse if they were organised into subdirectories by user and problem.

Would it be useful to create separate download links for just scoreboard.csv (and unofficial_scoreboard.csv)? (Most of the files in the zip file are machine-readable and not user friendly, while the scoreboard CSV is pretty much the opposite; so to me it makes sense to separate them.)

I'm curious about the commit 1d4e205e35dc45 "Disallow organisers from exporting contests", what's the reasoning behind that?

app/exporters/contests/contest_exporter.rb
```diff
@@ -0,0 +1,100 @@
...
+      problems = contest.problem_set.problems
+
+      scoreboard = CSV.generate do |csv|
+        csv << ["Rank", "User ID", "Real Name", "Username", "School", "Year"] + problems.map{ |p| p.name } + ["Total Score", "Time"]
+
+        num_ranked = 0
+        rank = nil
+        previous_record = nil
+
+        contest.scoreboard.each do |record|
```
bagedevimo, line 18: i think if you do a .includes(user: :school) you'll skip an N+1 accessing users and schools down in line 30

app/exporters/contests/contest_exporter.rb
```diff
@@ -0,0 +1,100 @@
...
+          row += [record.score, format("%d:%02d:%02d",record.time_taken.to_i/3600,record.time_taken.to_i/60%60,record.time_taken.to_i%60)]
+          csv << row
+        end
+      end
+    end
+
+    def export(path, options = {})
+      contest = subject
+      subpath = File.expand_path('submissions', path)
+      dir.mkdir(subpath)
```
bagedevimo, line 43: i may be totally wrong but i thought `dir` had to be `Dir`

app/exporters/contests/contest_exporter.rb
bagedevimo: hmm that might be a rails 4.2 feature actually..

app/exporters/contests/contest_exporter.rb
bagedevimo: and a few instances of `file` -> `File` ?

app/exporters/contests/contest_exporter.rb
tom93: `scoreboard` is a custom method, not an ActiveRecord relation. It already has `.includes(:user)` inside it, but that might not load the schools... Should investigate if it's making lots of queries for the schools and if `.includes(user: :school)` fixes that.

app/exporters/contests/contest_exporter.rb
tom93: It's doing fancy stuff with zip files, see base_exporter.rb.
For context: The code in this PR has already been pushed to the server and works.

app/exporters/contests/contest_exporter.rb
bagedevimo: ah, cool, i see the sitch now. ty for the learning :)

app/exporters/contests/contest_exporter.rb
bagedevimo:
```      self.dir = Dir
      self.file = File
```

Haha this is pretty unfair on me.

bagedevimo: okay, had a good read - this is the most "C" like ruby i've ever seen. be worth making it a little more ruby-like down the road, but not an issue for now :)

Holmes98:
Thanks for reviewing. The proposed changes in add-contest-exporter-v3c-efficient-submissions-export-wip all look reasonable to me; I can try testing them in the weekend.

This feature is pretty much only being used for NZIC, as far as I'm aware. The JSON files are intended to provide a dump of all the contest metadata that could be used for various scripts (e.g. to generate statistics, detect cheating), but they do include some sensitive data (full names, email addresses) which organisers don't have access to, which was the reason for 1d4e205e35dc45.

We already have a few scripts that work with the current format so I'd prefer to leave them as is, but I wouldn't mind if we changed the directory structure for submissions to `/{user_id}/{problem_id}/{submission_id}.{extension}` or something similar. I also wouldn't be opposed to having extra download links for just the scoreboards, but it wouldn't make a difference for NZIC since we'd always do a full export.

tom93 (approved):
I tested my proposed changes, fixed a typo (5be2eb06822c50fbe4cef66972d7ce736e63708a), then rebased and pushed to this PR's branch.

I'm approving this version, and will use separate PRs for the bigger changes.

I propose to squash-and-merge with commit message "Add contest exporter (#142)".

---

> We already have a few scripts that work with the current format so I'd prefer to leave them as is

The biggest issue I have with the current representation is how the individual scores are stored (details below). I'd very much like to change that (in a separate PR). The scripts will need to be tweaked a little, I hope that's not a problem; I'd be happy to help.

Currently the ContestScore fields (which are per-problem-per-user) are embedded in the parent ContestRelation object (which is per-user), with the problem ID appended to each field name. Example (pseudo-JSON, with two problems whose IDs are 111 and 222):

```
scoreboard: [
    {
        user_id, score, time_taken, school_id,
        score_111, attempt_111, attempts_111, sub_111,
        score_222, attempt_222, attempts_222, sub_222
    },
    {
        user_id, ...
    }
]
```

I think it would be much nicer (both for the app's code and for the scripts) if the ContestScore objects were stored in an array, in the same order as the problems:

```
scoreboard: [
    {
        user_id, score, time_taken, school_id,
        scores: [
            { score, attempt, attempts, sub },
            { score, attempt, attempts, sub }
        ]
    },
    {
        user_id, ...
    }
]
```

Such an array can be produced using the following snippet:

```ruby
contest_relation.contest_scores.index_by(&:problem_id).values_at(*problem_ids)
```

Another option is to store them in a hash, with the problem ID as the key. This corresponds very closely to the old representation.

Yet another option (which I really don't like) is to use an unordered array of ContestScore objects. I don't like it because it is awkward to use from scripts, but it is worth mentioning because it directly corresponds to the existing `contest_scores` association.

Holmes98:
> The biggest issue I have with the current representation is how the individual scores are stored (details below). I'd very much like to change that (in a separate PR).

Oh sure, feel free to make changes to anything under the contest JSON. I thought it could be useful at some point but we're currently only using the submissions JSON.

tom93:
Shall I deploy? (Hard-reset etc.)

Holmes98:
Sure

tom93:
Done!

Issue #144, opened 2022-06-21
SMTP authentication failure when the app sends emails for account confirmation etc.


tom93:
**Status:** A temporary fix has been implemented and deployed.

A user has reported that sign up form no longer works (it produces a generic error message).

The following extract from *log/production.log* shows it is due to an SMTP authentication failure when sending the confirmation email from our @gmail.com account:

```
Net::SMTPAuthenticationError (535-5.7.8 Username and Password not accepted. Learn more at
):
  app/controllers/accounts/registrations_controller.rb:6:in `create'
  lib/rack/response_timer.rb:12:in `_call'
  lib/rack/response_timer.rb:7:in `call'
```

(The SMTP response is cut off, but apparently just links to https://support.google.com/mail/?p=BadCredentials which isn't very helpful.)

The problem is that Google recently stopped allowing username-and-password authentication for third-party apps (the "Allow less secure apps" option):

> To help keep your account secure, from May 30, 2022, Google no longer supports the use of third-party apps or devices which ask you to sign in to your Google Account using only your username and password.  
> —https://support.google.com/accounts/answer/6010255

I implemented and deployed a temporary workaround in the branch 'hotfix-email' ([commits](https://github.com/NZOI/nztrain/compare/743679784bebc5bc2a37e7e10d83bef6d033395d...a54b2ca6a61344cfe30c79893ed678c72efb3eb9)). The workaround is to write the emails to a file and ask users to contact us so we can manually send them the confirmation link.

To fix this properly we plan to enable 2-step verification for the app's Gmail account, which makes it possible to generate an app-specific password (https://support.google.com/accounts/answer/185833). Authentication using an app-specific password is still supported. This process is currently delayed because the account's recovery details need to be updated, and [this can take a week](https://support.google.com/accounts/answer/3463280).

Another approach is to switch to OAuth for authentication, but that's complicated.

tom93:
Status update:
We are still working on enabling 2-step verification.
Meanwhile, I implemented an improved version of the temporary workaround in #145 (not deployed yet).

tom93:
This has now been fixed properly by enabling 2-step verification and generating an app-specific password.
I reverted the temporary workarounds and verified that mail delivery works.

PR #145, opened 2022-06-23
Hotfix: Deliver emails to a Discord channel using a webhook

tom93:
Improved workaround for #144.

I deployed the initial workaround directly (branch 'hotfix-email', [commits](https://github.com/NZOI/nztrain/compare/743679784bebc5bc2a37e7e10d83bef6d033395d...a54b2ca6a61344cfe30c79893ed678c72efb3eb9)), but I decided to use a pull request for this improved version to give people a chance to comment, because it's a larger change and the initial workaround has bought us time.

I intend to deploy this in ~24 hours if there are no objections.

We are pretty close to enabling 2-step verification (need the right people to be online at the same time), so we could just wait a bit and not deploy this workaround.

Another option is to keep the initial workaround (file delivery) but change the instructions in the web interface to only ask students to contact staff if they don't get the email within a day (currently the instructions ask students to contact staff unconditionally). I'm happy to check the mail directory on the server at least once a day, so we can still get a reasonable response time without using Discord and without asking the students to send us an email.

---

Commit message:

> Deliver emails to a Discord channel using a webhook
> 
> This is nicer than writing the emails as files because we can receive
> notifications from Discord (so students shouldn't need to send us a
> request) and because staff will be able to view and forward the emails
> without requiring SSH access to the server.
> 
> The webhook URL can be created in Discord by opening the Server
> Settings then the Integrations tab:
> https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks
> 
> This URL needs to be set in the web app using admin > settings > New
> *before* this commit is deployed. The key is "system/mailer/webhook"
> and the value is the webhook URL.
> 
> This commit also updates the messages in the interface to only ask
> students to contact staff if they don't get the email within a day.
> 
> ...

Here is a sample message:

> To: example@example.com
> Subject: Confirmation instructions
> 
> Welcome example@example.com!
> 
> You can confirm your account email through the link below:
> 
> Confirm my account: http://train.nzoi.org.nz/accounts/confirmation?confirmation_token=xxx
> .

The idea is that the staff on Discord will be notified and will manually email this message.

You can see my development history (bugfixing) in branch 'tom93/hotfix-email-v2-history' ([commits](https://github.com/NZOI/nztrain/compare/a54b2ca6a61344cfe30c79893ed678c72efb3eb9...da7af74681a9772702abedfc8ed279541658c8f3)).

### DIFF
--- a/lib/webhook_mail_delivery.rb
+++ b/lib/webhook_mail_delivery.rb
@@ -1,0 +1,60 @@
+require 'uri'
+require 'net/http'
+require 'nokogiri'
+
+# WebhookMailDelivery posts emails using a webhook, e.g. to a Discord channel.
+#
+# It allows staff to then manually send the email, which is useful when regular
+# delivery using SMTP is broken.
+#
+# Special characters in the message are backslash-escaped (this is suitable for
+# destinations that render using Markdown-like syntax, such as Discord).
+class WebhookMailDelivery
+  attr_accessor :settings
+
+  def initialize(values)
+    self.settings = values
+  end
+
+  def deliver!(mail)
+    body_html = mail.body.decoded
+    body_text = html_to_plain_text(body_html)
+    body_text = body_text.rstrip + "\n.\n" # add a visual separator between messages
+    body_escaped = escape_message_for_markdown(body_text)
+    post_webhook "To: #{mail.to.join(' ')}\nSubject: #{mail.subject}\n\n#{body_escaped}"
+  end
+
+  def html_to_plain_text(html)
+    # convert links to plain text
+    # e.g. '<a href="https://...">Click here</a>' becomes 'Click here: https://...'
+    html = html.gsub(/<a href="([^"]*)">([^<]*)<\/a>/, '\2: \1')
+    # strip remaining HTML tags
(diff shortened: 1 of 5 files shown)


lib/webhook_mail_delivery.rb
```diff
@@ -0,0 +1,60 @@
...
+    Nokogiri::HTML(html).text
+  end
+
+  def escape_message_for_markdown(text)
+    # replace non-ASCII characters with "?"
+    text = text.encode('ascii', :undef => :replace, :invalid => :replace)
+    # replace control characters (other than newline, \x0A) with "?"
+    text = text.gsub(/[\x00-\x09\x0B-\x1F\x7F]/, '?')
+    # backslash-escape all characters except those known to be safe (very conservative)
+    text = text.gsub(/[^A-Za-z0-9_ \n]/) {|s| '\\' + s}
```
tom93, line 41: This should have been

```
     text = text.gsub(/[^A-Za-z0-9 \n]/) {|s| '\\' + s}
```

because underscores need to be escaped (done commit db3bab51fa734586092233ed5f134909559182f1).

PR #146, opened 2023-01-11
Give group members access to scoreboard

puqeko:
It was discussed at the AGM that it would be motivating for the training squad to see the progress of their pairs. Admins have access to a scoreboard which offers exactly this. This pull request seeks to give group members permission to see the group scoreboard.

* These changes are untested (I don't have a VM setup to do that).
* This means all members of a group can see the progress of all other members in that group. This would apply to all groups.

Would be grateful if someone could please review this.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -73,8 +73,8 @@ class GroupsController < ApplicationController
       format.html { render :layout => "group" }
     end
   end
-  
-  def info 
+
+  def info
     @group = Group.find(params[:id])
     authorize @group, :show?
     respond_to do |format|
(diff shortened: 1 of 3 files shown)


tom93 (approved):
Hi @puqeko!

I tested and pushed some changes. Approving, but I'd like to review & merge #142 first (since it has already been deployed).

Regarding the submission counts, I agree with hiding them on the group scoreboards for non-admins. I note that the *contest* scoreboards show submission counts to everyone, but I think the privacy expectations are different when practising as opposed to competing.

Additionally, the group scoreboard has a subtle feature that indicates whether a user has every viewed each problem. I think it is privacy-intrusive to show this information to other group members, so I went ahead and pushed a commit that disables this feature for non-admins (all cells are displayed as if the user viewed the problem).

For merging, I suggest squash-and-merge with the following commit message:

```
Give group members access to group scoreboard
```

```
The group scoreboard page shows the scores of members on all the
problems in the group. Previously it was only visible to admins. This
commit gives group members access to this page because we hope it
would be motivating for students to see the progress of their peers.

The group scoreboard has a subtle feature where a completely blank
cell indicates that the user has never viewed the problem. This commit
disables this feature for non-admins, to give users some privacy from
other group members.

This commit also hides the submission counts when the group scoreboard
is viewed by a non-admin, again to give some privacy. (The existing
contest scoreboards show the submission counts, but the privacy
expectations may be different when practising as opposed to
competing.)

Co-authored-by: puqeko <puqeko@gmail.com>
Co-authored-by: Tom Levy <tomlevy93@gmail.com>
```

PR #149, opened 2023-02-07
Upgrade rmagick

bagedevimo:
This newer version workers with older system libraries (as far as I can tell) and workers with new ruby / rails versions so is a requirement for eventually upgrading Rails

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,6 @@ gem 'forem-redcarpet', github: 'NZOI/forem-redcarpet'
 
 gem "nokogiri", '~> 1.10.8'
 gem 'redcarpet'
-#gem 'rmagick', '2.13.2'
 gem 'rmagick'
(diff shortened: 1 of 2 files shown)


tom93:
Haven't tested yet, but here is a suggested commit message (spelling):

```
Upgrade rmagick
```

```
This newer version works with older system libraries (as far as I can
tell) and works with new Ruby / Rails versions so is a requirement for
eventually upgrading Rails.
```

bagedevimo:
rebased and fixed up the commit as you suggested, thanks

tom93 (approved):
Approved (but merge #163 first).
I (re-)pushed a small fix to remove an old commented-out version in Gemfile.

GitHub's squash-and-merge doesn't support the "fixup!" notation, so the commit message will have to be cleaned up manually.

PR #150, opened 2023-02-07
Draft: Create rubyonrails.yml

bagedevimo:


tom93:
We are considering to remove maxmind, see #108.

bagedevimo:
> We are considering to remove maxmind, see #108.

I love the idea of dropping it but I'm going to try get this working with as fewer code changes as possible, because then I can be more confident about future changes like dropping maxmind knowing that CI has my back :)

tom93:
> I love the idea of dropping it but I'm going to try get this working with
> as fewer code changes as possible, because then I can be more confident
> about future changes like dropping maxmind knowing that CI has my back :)
>
Okay, but beware that maxmind is currently broken (breaks sign up form --
see linked issue; CI should still pass because the specs don't cover
that...)

Also, see #82 for an earlier PR for using GitHub Actions (I haven't
reviewed it yet).

PR #160, opened 2023-02-09
Fix divide-by-zero in ApplicationHelper.progress_bar when weighting is 0

tom93:
The first patch is a code change that was made directly on the production server, I'm not sure who originally wrote it or under what conditions the weighting is 0.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -59,14 +59,17 @@ module ApplicationHelper
       text = link_to(text,link,{:class => "progress_link", :style => "display: block; width: 100%;"})
     end
     weighting = 100 if weighting.nil?
-    percent = score*100/weighting unless score.nil? || weighting == 0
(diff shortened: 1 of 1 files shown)


bagedevimo (dismissed):

Holmes98:
Just FYI, I applied that hotfix since we changed the NZIC welcome problem to be weighted at 0 and it was breaking the contest page for active contestants (I only found out during the contest). I'd like to look into a better solution, since it currently always looks like the problem is unsolved, but that can wait for a separate PR.

bagedevimo:
yeah, i figured that, but i'm very pro just getting master up to date with what is actually deployed and we can revisit as needed. we should also address the "raw" call as they're generally considered a security risk, and i think we can do this entire implementation using an html <progress> element.

Holmes98 (approved):
@tom93 I'll let you merge if you're fine with this?

PR #159, opened 2023-02-08
Drop references the NZIC module, remove the NZIC models from the bundle

bagedevimo:
These appear to be unused, and are certainly broken. Commit messages don't paint a clear picture of why they are there and they have gone a long time since they were last actually edited in a meaningful way.

Closes #152.

PR #161, opened 2023-02-10
Drop forem, replace with placeholder page to avoid dead links

### LINKED ISSUE #33: Remove the forum
NZTrain has a forum, but it's largely empty, and the library that provides it has been unmaintained for years.

Even if we need a forum in the future, it'll have little code in common with what's here now.

Let's remove it.

### LINKED ISSUE #153: Drop forem
The forums have a message saying they're unused, but folks are obviously still trying to post (5 months ago last attempt) so it might be best to remove the menu item reference to the community, and replace the link (`/forum`) with a static page indicating where other NZOI communities hang, etc.

Removes a huge number of gems and dependencies from the application, that were never *really* used aren't aren't being used now

bagedevimo:
Closes #33 
Closes #153


bagedevimo:
Merge #159 first

Gemfile
```diff
@@ -40,7 +34,6 @@ gem 'simple_form', '3.0.1'
 gem 'facebox-rails'
 gem 'strong_presenter', '~> 0.2.2'
 gem 'render_anywhere'
-gem 'pygments.rb', '0.5.4'
```
tom93, line 43: We use pygments.rb in the app (to do syntax highlighting on submissions), so we can't remove this dependency.

tom93 (changes requested):

Gemfile
bagedevimo: yep you're right - did the agressive thing to start with, but i've resolved that now and removed the WIP: tag, so I think this is good review :)

bagedevimo:
Rebased on to master and force-pushed over top, so you may need to do a fetch and reset hard sorry Tom.

tom93:
Do we want to take some sort of backup for the forums? (E.g. an offline copy, or a static public archive.)

> Rebased on to master and force-pushed over top, so you may need to do a fetch and reset hard sorry Tom.

No worries, I haven't even committed anything on that branch.

bagedevimo:
>Do we want to take some sort of backup for the forums? (E.g. an offline copy, or a static public archive

Yeah I've intentionally not bothered to drop the tables so we can have this discussion first. You've gotta remove references to tables before dropping anyway, but I was going to ask one of you to pg_dump some of the tables before I drop them in a follow up migration series.

config/routes.rb
```diff
@@ -300,4 +292,6 @@
 
   # See how all your routes lay out with "rake routes"
 
+  # Legacy routes maintained to avoid routes breaking
+  get "/forum/*anything" => "pages#forum"
```
tom93, line 296: Nit: This doesn't match /forum or /forum/, no big deal.

tom93 (dismissed):
I took a snapshot of the forum's HTML (there were only 17 posts) and put it in the NZOI Google Drive folder, so this is good to go.

config/routes.rb
bagedevimo: hmm thats annoying tho - i'll fix it. the entire point is to avoid dead SEO links or bookmarks, as unlikely as they may be.

bagedevimo: thanks for catching this - have fixed and added a spec

tom93 (changes requested):
Oops, just realised this breaks `rake db:migrate` (which we do in fresh installs, see script/install/migrate.bash; probably better and faster to use `rake db:setup`, but that's a discussion for another time).
What should we do? Drop the forem migrations?

Thanks for fixing the redirect and adding the spec. I pushed a tiny commit to make the example more realistic :)

tom93 (dismissed):
My db/schema.rb was slightly different after running `bundle exec rake db:drop db:create db:migrate` -- some more forem tables disappeared. I pushed a commit  with my schema.rb (63ad0a26952940111e713cbb92ee3cf50bcfa431, all references to forem completely gone). Can I merge? (Preferably after #171 and #172 to keep the XML commits together)

bagedevimo:
Merge at your leisure :)

tom93 (approved):

tom93:
The DB tables and columns were manually dropped by temporarily fixing the forem migrations to be reversible (commit 98d3aec13272614315182a78d75b83585f000922) and then rolling them back using

```
for i in $(cd db/migrate && ls -r *forem*); do
   RAILS_ENV=production bundle exec rake db:migrate:down VERSION=$i || break
done
```

PR #162, opened 2023-02-10
Switch to the default CookieJar session store

bagedevimo:
Deploying this will log everyone out, so we may want to consider an intermediate migration step.


bagedevimo:
Merge #159 first

bagedevimo:
Tested by logging in ticking remember me, stopping the rails server, exiting my browser. change to this branch, start rails, start browser and verify still logged in. The cookie works fine for the non-session remember me, but if you don't click remember me your session will be lost regardless of the browser, etc.

tom93:
Merge #195 first, to skip the legacy signed-but-not-encrypted cookie store mode.

Holmes98 (approved):

tom93 (approved):
We decided over Discord not to bother with a migration or a notice about being signed out (if they didn't tick "remember me").

tom93:
This has been deployed just now, so all users that didn't tick "remember me" will be signed out.

tom93:
It appears that users might get signed out even if they ticked "remember me", we are not quite sure why.
Users that have been signed out should be able to sign in again using their password as usual.
If you forgot your password, you can use the "Forgot your password?" option to reset it.
Sorry for any hassle.

PR #164, opened 2023-02-11
Bump pygments.rb from 0.5.4 to 1.1.2 [Ruby 2.4]

tom93:
I'm not sure when we'll try to upgrade to Ruby 2.4 (probably not any time soon because we'll first need to [upgrade to Rails 4.2](https://rubyonrails.org/2017/2/21/Rails-4-2-8-has-been-released)), but I happened to came across this particular dependency issue and had a go fixing it (that before I found out how difficult it would be to upgrade to Ruby 2.4), so I thought I might as well create a PR.

Note that forem-redcarpet also depends on pygments.rb 0.5.4, so before we merge this PR we need to either
 - update that dependency in our existing fork of forem-redcarpet ([NZOI/forem-redcarpet#1](https://github.com/NZOI/forem-redcarpet/pull/1)), or
 - remove forem ([#161](https://github.com/NZOI/nztrain/pull/161)).

From the commit message:

> yajl-ruby < 1.3.0 doesn't support Ruby 2.4, see [1].
> pygments.rb added support for yajl-ruby 1.3 in version 1.0.0, see [2].
> 
> But also, yajl-ruby <= 1.4.2 has security vulnerability.[3]
> pygments.rb 1.1.0 removed the dependency on yajl-ruby.[4]
> 
> So update pygments.rb to >= 1.1.0.
> 
> Note that pygments.rb 2.0.0 has some breaking changes relating to timeout handling[5], so I am keeping the update minimal for now.
> 
> [1]: https://www.github.com/brianmario/yajl-ruby/pull/165
> [2]: https://www.github.com/pygments/pygments.rb/pull/164
> [3]: https://www.github.com/brianmario/yajl-ruby/security/advisories/GHSA-jj47-x69x-mxrm
> [4]: https://www.github.com/pygments/pygments.rb/pull/170
> [5]: https://github.com/pygments/pygments.rb/blob/master/CHANGELOG.adoc#200rc1-2021-01-07---slonopotamus


bagedevimo (approved):

tom93:
I didn't test this well enough... this update causes a small glitch with the line numbers (e.g. when viewing submissions). The numbers "1" and "2" appear next to the first line on top of each other:
submission-screenshot
I'm looking into this, please don't deploy yet (thought it's not such a big deal).
So far it looks like the extra line number is caused by a new empty `<span>` element being generated at the start.

bagedevimo:
almost certainly gonna be a css glitch - can you drop me a quick repro set of steps and i'll take a peek?

tom93:
Should just have to create a problem and submit.
Looks like it's our CSS (app/assets/stylesheets/code.css) which needs to be updated for the new generated HTML.

tom93:
This regression has been reported to Pygments but won't be fixed: https://www.github.com/pygments/pygments/issues/961

bagedevimo:
What a PITA. Do you wanna update the CSS or would you like me too? I'm happy either way.

tom93:
I can update the CSS. I'm thinking of changing the selector from

```css
pre > span::before {
```

to

```css
pre > span[id^="line-"]::before {
```

(i.e. only consider <span> elements that have an id attributes that starts with "line-"). This seems to work fine.

Here an example of the old HTML (pygments.rb 0.6.3):

```html
<div class="highlight"><pre><span id="line-1"><span class="n">foo</span>
</span><span id="line-2"><span class="n">bar</span>
</span></pre></div>
```

And the new HTML (pygments.rb 1.0.0):

```html
<div class="highlight"><pre><span></span><span id="line-1"><span class="n">foo</span>
</span><span id="line-2"><span class="n">bar</span>
</span></pre></div>
```

PR #165, opened 2023-02-12
Refactor to remove the UserPresenter service object

bagedevimo:
This is a pretty thin wrapper around the User, providing only some helpers for permissions. We *probably* have to write out the strong presenters library due to rails upgrade / ruby upgrade reasons, so this is a good place to start.


tom93 (approved):
I changed caps ("Real name" back to "Real Name") for consistency with other fields, but kept "E-mail" => "Email" (because we use no-hyphen everywhere else).

(Aside: I tried using an empty commit message so GitHub's squash-and-merge won't include it; mostly works but GitHub still munges the original commit message. Probably won't do it again because empty commit messages are weird.)

PR #171, opened 2023-02-17
Add an `after_filter` to assert that XML is forbidden

tom93:
Commit message:
 
> The `content_type=` check implemented in #169 runs fairly early (which is good) but it is slightly fragile (e.g. it is bypassed if `response.content_type=` is called directly).
> 
> Add an `after_filter` that raises an error if XML slipped through.

I'm not sure what's the idiomatic behaviour for such assertion failures, so I just used `raise "..."` (which [raises a RuntimeError](https://ruby-doc.org/3.2.1/Kernel.html#method-i-raise)).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,6 +14,7 @@ class ApplicationController < ActionController::Base
   before_filter :read_settings
   before_filter :check_su_loss
   before_filter :configure_permitted_parameters, if: :devise_controller?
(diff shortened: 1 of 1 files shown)


Holmes98 (approved):

bagedevimo (approved):
This is fine - we should definitely "just" rewrite these methods to seperate controller / actions down the road but this is good for now.

PR #177, opened 2023-02-25
Remove the AI Contest implementation

bagedevimo:
Currently unused, a lot of code to upgrade so we'll bring it back later after some more thought about how exactly to maintain it.


tom93 (approved):
I pushed a commit (96762cfd21774a3d3082d63831c224d8f0d9e745) to also delete migrations (because they break fresh installs), views, assets, and commented-out routes (or did you leave them there on purpose?)

Should we drop or keep the DB tables? They contain one problem with several submissions. It would be a shame to lose that problem, so if we drop the DB tables I think we should try to export the problem and back it up somewhere (in a nicer format than a DB backup). I'm not sure whether the submissions are worth keeping. (Maybe create a backup of the AI Contest tables, and also a nice export of the problem?)

Suggested commit message body (the last part assumes we want to drop the DB tables):

~~~
The idea of the AI Contest is to have two programs compete against
each other in a game. This code is currently unused and it's difficult
to get it working.

We have a lot of other code to upgrade, so we are removing the AI
Contest for now. We might bring it back later after some more thought
about how exactly to maintain it.

Before deploying this commit, the migrations should be brought down
using the following command, which will drop the AI Contest tables:

```
for i in $(cd db/migrate; ls -r *_ai_* 20130113010505_add_uniq_index_to_games.rb); do
    bundle exec rake db:migrate:down VERSION=$i || break
done
```
~~~

bagedevimo:
Your changes on this are :+1: from me, obviously I can't review it now.

bagedevimo:
IMO a database backup is sufficient to retain this data - we can easily read it out of the backup locally using psql :)

Holmes98 (approved):

PR #180, opened 2023-03-02
Allow redis config to use ERB variables

bagedevimo:
Makes dockerisation much easier as we can embed a version of this file in the image that is looking for specific env vars. More recent rails versions ditch the .yml.template concept for just providing more easily customisable .yml's by default. This will move us towards doing this, also.

### DIFF
--- a/config/initializers/redis.rb
+++ b/config/initializers/redis.rb
@@ -1,6 +1,6 @@
 require 'redis'
 
-config = YAML.load( File.open( Rails.root.join("config/redis.yml") ) ).symbolize_keys
+config = YAML.load(ERB.new(File.read(Rails.root.join("config/redis.yml"))).result).symbolize_keys
 REDIS_CONFIG = config[:default].symbolize_keys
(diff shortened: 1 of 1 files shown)


tom93 (approved):
I force-pushed to rewrite the commit message (I'm pretty sure the .yml.template pattern is specific to our install scripts, and I think it's worth mentioning that Rail's database.yml supports ERB by default).

[Moved to #184] In the future, I think we should also:
1\. Allow ERB in all the other config/*.yml files
2\. Add a dependency on [dotenv-rails](https://rubygems.org/gems/dotenv-rails)
3a. Create a skeleton .env file with all the supported environment variables
3b. Replace the *.yml.template files with *.yml files that use ERB to look up the appropriate environment variables, with fallback to sensible defaults
3c. Remove (most of) script/install/config.bash and script/install/nztrain.bash

bagedevimo:
That approach you've outlined is perfect IMO!

PR #181, opened 2023-03-02
Avoid a local jump error caused by trying to return from a block

bagedevimo:
Returning from a block isn't really supported - instead you're supposed to use break / next to early return from a block. In any rate, this callback just needs to return true, which we can do via implicit return.

Rails 4.1 treats a non-local jump in a ActiveRecord::Callback as an error, so we have to fix this before.


tom93 (approved):
I force-pushed to add a note about how Rails 5.1 deprecates the return value, to avoid confusion in the future.

(I checked that we don't have any other returns outside of methods.)

PR #182, opened 2023-03-02
Introduce a temporary model for older migrations

bagedevimo:
Remove data changes in older migrations.

Means you can migrate from zero -> now without any issues. Only the
"data only" aspects of the migrations were changed, preserving all
schema changes. This ensures we can migrate from start to finish
cleanly, and that our migrations aren't relying on schema or data that
won't be present in a clean database.

More generally, this is why people introduce tools like data-migrate [1]
and avoid using models in the their regular migrations, but hey.

### DIFF
--- a/db/migrate/20131209063832_fix_bad_markdown.rb
+++ b/db/migrate/20131209063832_fix_bad_markdown.rb
@@ -1,21 +1,7 @@
 class FixBadMarkdown < ActiveRecord::Migration
-  def fix_titles(string)
-    toplevel = 10
-    pos = 0
-    while inc = (string.slice(pos, string.length) =~ /^((#+)[^#].*)$/)
-      toplevel = [toplevel, $~[2].length].min
-      pos += inc + $~[0].length
-    end
-    toplevel = 2 if toplevel<2
-    string = string.gsub(/^#{Array.new(toplevel-2,"#").join}/,'').gsub(/^(#+)([^# ])/, '\1 \2')
-    return string
-  end
-
   def up
-    Problem.all.each do |problem|
-      problem.statement = fix_titles(problem.statement)
-      problem.save
-    end
+    # Data-only, legacy migration contents removed in order
+    # to maintain the ability to cleanly migrate
   end
 
   def down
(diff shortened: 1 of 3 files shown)


db/migrate/20131209063832_fix_bad_markdown.rb
```diff
@@ -1,4 +1,10 @@
 class FixBadMarkdown < ActiveRecord::Migration
+  class LegacyProblem < ActiveRecord::Base
+    self.primary_key = "problem_set_id"
+
+    self.table_name = "problem_sets_problems"
+  end
```
tom93, line 6: I'm pretty sure that in this migration it should just be `class Problem < ActiveRecord::Base; end`.

I haven't managed to test the other migration yet (need to do it when there are problems in the DB).

But wouldn't it be much simpler to just comment out the code, since our production DB is already shipshape and we don't have other DBs that need to be migrated?

I spotted another migration that may be problematic: 20131217112526_change_test_set_and_test_case_sample_and_prerequisites.rb

tom93 (changes requested):

bagedevimo:
Force pushed over the old PR with this new approach

bagedevimo:
Closing in favour of #187

tom93 (commented):
I dug into this and the root cause turns out to be an issue with ranked-model versions 0.4.3 - 0.4.4, so if we just avoid those versions (#187) we don't have to change our migrations.

There is definitely a wider problem with our data migrations and how they use the models (several migrations only work if the database is empty). Fixing the 2-3 cases that came up here because of ranked-model feels a little pointless, but I'm fine with merging anyway.

My only comment is that for 20131206074948_name_all_test_sets_and_cases.rb, I'd prefer to comment out the code instead of removing it, because I think the method `name_uniquely` is neat.

bagedevimo:
> My only comment is that for 20131206074948_name_all_test_sets_and_cases.rb, I'd prefer to comment out the code instead of removing it, because I think the method name_uniquely is neat.

Usually I rely on git for this, but your concern is that new contributors won't know to look for it?

tom93:
> > My only comment is that for 20131206074948_name_all_test_sets_and_cases.rb, I'd prefer to comment out the code instead of removing it, because I think the method name_uniquely is neat.
> 
> Usually I rely on git for this, but your concern is that new contributors won't know to look for it?

Yeah. (I wrote similar code for my personal helper functions, and when I reviewd this PR it was a fun surprise to see that this already existed. Not exactly discoverable, but still better than buried in the commit history.)

PR #183, opened 2023-03-02
Refactor calls to method_missing find_or_create_by_*

bagedevimo:
This is softly deprecated in Rails 4.0 [1] and are fully removed under Rails 4.1, so they need to be replaced before we can upgrade.

[1] https://guides.rubyonrails.org/4_0_release_notes.html#active-record-deprecations


tom93 (approved):
(I agree with switching to the "!" version)

tom93 (approved):

PR #189, opened 2023-03-10
Remove custom "distinct" scope, since it's provided by Rails, and Rails 4.1 forbids that name

tom93:
(Extracted from commit 04b1c7cf56f5d74b8c02a164f5cfac4ca79223fe of #188)

The "distinct" method is available since Rails 4.0.0 [1] (and was available as "uniq" since Rails 3.2.0 [2]), so we no longer need to define our own "distinct" scope.

And as of Rails 4.1 it is forbidden to define a scope with a conflicting name[3], so we have to remove our "distinct" scope.

[1] https://github.com/rails/rails/commit/a1bb6c8b06db83546179175b9b2dde7912c86f9b
[2] https://github.com/rails/rails/commit/562583c7667f508493ab8c5b1a4215087fafd22d
[3] https://www.github.com/rails/rails/pull/13450

### DIFF
--- a/app/models/contest.rb
+++ b/app/models/contest.rb
@@ -43,9 +43,6 @@ class Contest < ActiveRecord::Base
     end
   end
 
-  # Scopes
-  scope :distinct, -> { select("distinct(contests.id), contests.*") }
-
   def self.user_currently_in(user_id)
     joins(:contest_relations).where(:contest_relations => { :user_id => user_id }).where("contest_relations.started_at <= :time AND contest_relations.finish_at > :time",{:time => DateTime.now})
   end
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -25,9 +25,6 @@ class Group < ActiveRecord::Base
 
   validates :name, :presence => true, :uniqueness => { :case_sensitive => false }
 
-  # Scopes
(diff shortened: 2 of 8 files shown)


tom93:
*[The only place in our code that calls "distinct" is Problems.by_group, which is unused, so the following is moot.]*

I just noticed an oddity: Rails' "distinct" method behaves like standard SQL DISTINCT (de-dups based on identical result rows, not based on primary ID). So for example, `Problem.by_group(0).pluck(:statement)` is _not_ the same as `Problem.by_group(0).map(&:statement)`: the second query is intuitive and returns the statements of the distinct problems, while the first query returns the distinct statements (so if two problems have the same statement, the statement will only be included once).

Our custom scope behaves even more weirdly with "pluck". It uses `select("distinct(problems.id), ...")`, which is completely disregarded when "pluck" is used (so the results may multiple rows for a single problem).

It turns out that the Rails method has higher priority that our custom scope, and we have been using it unwittingly since we upgraded to Rails 4.0. So this PR won't change any behaviour.

Personally I find both behaviours surprising, but I suppose it makes more sense to adopt Rails' behaviour for consistency. Then the workaround is to include the ID in the results: `.pluck(:id, :statement)`.

Holmes98 (approved):

PR #190, opened 2023-03-10
Downgrade json to 1.8.6 for Rails 4.1 compatibility; add workaround for CVE-2020-10663

tom93:
Rails' activesupport 4.1.* depends on json ~> 1.7 (i.e < 2.0) [1], so in order to upgrade to Rails 4.1 we need to downgrade json from 2.6.3 to 1.8.6 (the highest < 2.0).

But json < 2.3 has a security vulnerability, CVE-2020-10663 [2]. So we also add a dependency on the gem 'json_cve_2020_10663', which monkey-patches the json gem to protect from this vulnerability.

(Verified that the CVE is patched using the code listed on [3].)

[1] https://rubygems.org/gems/activesupport/versions/4.1.16
[2] https://www.ruby-lang.org/en/news/2020/03/19/json-dos-cve-2020-10663/
[3] https://github.com/rails-lts/json_cve_2020_10663

Suggested-by: Jonathan Khoo <jkhoo98@gmail.com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby '>= 2.3.8', '< 2.4' # must be < 2.4 until we upgrade to Rails >= 4.2.8 (see https://rubyonrails.org/2017/2/21/Rails-4-2-8-has-been-released)
 
 gem 'rails', '~> 4.0.0'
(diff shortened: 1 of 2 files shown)


Holmes98 (approved):
I'd prefer a shorter commit summary (e.g. `Downgrade json to 1.8.6 for Rails 4.1 compatibility`), mostly so it looks nicer on GitHub, but it's not a big deal. Will leave it up to you.

PR #191, opened 2023-04-01
Fix contestants page

Holmes98:
It was using an implicit join reference, which was deprecated in Rails 4.0 and removed in Rails 4.1.

Currently throws the following error:
```
PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "users"
LINE 1: ...t_id" = $1  ORDER BY contest_relations.school_id, users.user...
                                                             ^
: SELECT "contest_relations".* FROM "contest_relations"  WHERE "contest_relations"."contest_id" = $1  ORDER BY contest_relations.school_id, users.username
```

Under Rails 4.0, it was logging this warning:
```
DEPRECATION WARNING: It looks like you are eager loading table(s) (one of: contest_relations, users) that are referenced in a string SQL snippet. For example: 

    Post.includes(:comments).where("comments.title = 'foo'")

Currently, Active Record recognizes the table in the string, and knows to JOIN the comments table to the query, rather than loading comments in a separate query. However, doing this without writing a full-blown SQL parser is inherently flawed. Since we don't want to write an SQL parser, we are removing this functionality. From now on, you must explicitly tell Active Record when you are referencing a table from a string:

    Post.includes(:comments).where("comments.title = 'foo'").references(:comments)

If you don't rely on implicit join references you can disable the feature entirely by setting `config.active_record.disable_implicit_join_references = true`. (called from block in _app_views_contests_contestants_html_erb___2774875097299588441_47270994716840 at /apps/nztrain/app/views/contests/contestants.html.erb:159)
```

### DIFF
--- a/app/controllers/contests_controller.rb
+++ b/app/controllers/contests_controller.rb
@@ -89,7 +89,7 @@ class ContestsController < ApplicationController
     @contest = Contest.find(params[:id])
     authorize @contest, :contestants?
     @groups = Group.all
-    @contest_relations = @contest.contest_relations.includes(:user).order("contest_relations.school_id, users.username")
+    @contest_relations = @contest.contest_relations.includes(:user).order("contest_relations.school_id, users.username").references(:user)
 
     render layout: 'contest'
   end
--- a/app/models/contest.rb
+++ b/app/models/contest.rb
@@ -38,7 +38,7 @@ class Contest < ActiveRecord::Base
     problems = problem_set.problems
     self.contest_relations.find_each do |relation|
       problems.each do |problem|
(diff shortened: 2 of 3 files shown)


Holmes98:
Also throws errors on updating extra time (and probably a few other cases), due to removal of dynamic finder methods.
Deprecated in 4.0: https://guides.rubyonrails.org/4_0_release_notes.html#active-record-deprecations
Removed in 4.1: https://github.com/rails/rails/blob/0cad778c2605a5204a05a9f1dbd3344e39f248d8/activerecord/CHANGELOG.md?plain=1#L2377

tom93 (approved):
(didn't test)

PR #193, opened 2023-12-20
Upgrade devise to 3.4.1

bagedevimo:
Devise 3.4.0 added support for Rails 4.2 so it's in the critical path for that set of upgrades. 3.4.1 fixed an issue with 3.4.0, so seems worth the the patch bump.

https://github.com/heartcombo/devise/blob/3-stable/CHANGELOG.md#341---2014-10-29

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ ruby '>= 2.3.8', '< 2.4' # must be < 2.4 until we upgrade to Rails >= 4.2.8 (see
 gem 'rails', '4.1.16'
 gem 'json_cve_2020_10663', '~> 1.0' # required until we update json >= 2.3, which we can only do once we upgrade to Rails >= 4.2 because activesupport 4.1.* depends on json ~> 1.7 (i.e < 2.0): https://rubygems.org/gems/activesupport/versions/4.1.16
 
-gem 'devise', '~> 3.2.2'
+gem 'devise', '~> 3.4.1'
 gem 'psych', '~> 2.0.2' # part of stdlib, need newer version for safe_load
 
 # change back to cookie-based store (encrypted)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -97,10 +97,11 @@ GEM
     crass (1.0.5)
     daemons (1.3.1)
     debug_inspector (0.0.3)
-    devise (3.2.4)
+    devise (3.4.1)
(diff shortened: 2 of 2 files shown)


tom93 (commented):
(I didn't test this.)
I noticed that the default messages have changed across this version bump, but we can update them in a separate PR (our config/locales/devise.en.yml is quite old, not sure if there is an automated way to update but I can do a manual three-way merge).

tom93 (approved):
~~(I didn't test this.)~~ *Edit: Tested, see below.*
I noticed that the default messages have changed across this version bump, but we can update them in a separate PR (our config/locales/devise.en.yml is quite old, not sure if there is an automated way to update but I can do a manual three-way merge).

tom93 (changes requested):
Hold on, for me it also updates bcrypt (3.1.11 => 3.1.20); investigating...

tom93 (approved):
When I bump the devise dependency in the Gemfile then run `bundle install` (or `bundle update --conservative devise`), it also bumps bcrypt, presumably because bundler tries updating transitive dependencies -- I don't think devise actually depends on a newer version of bcrypt (the devise changelog mentions they bumped the bcrypt dependency in 3.2.4, but we are already on 3.2.4 so it's not new). Not sure why that didn't happen for you, I have a weak preference to letting it bump (bcrypt changelog seems safe), but doesn't seem important.

Holmes98:
Worth noting that bcrypt >= 3.1.12 is required on Ubuntu 20.04 (aba727e), but that can be done separately.

bagedevimo:
> ~(I didn't test this.)~ _Edit: Tested, see below._ I noticed that the default messages have changed across this version bump, but we can update them in a separate PR (our config/locales/devise.en.yml is quite old, not sure if there is an automated way to update but I can do a manual three-way merge).

You are 100% correct, but I couldn't see a reason to do this in a hurry, and they do update these localisation files again in the next major version while we'll have to upgrade too soon-enough.

bagedevimo:
>Not sure why that didn't happen for you, I have a weak preference to letting it bump (bcrypt changelog seems safe), but doesn't seem important.

bundler version is my guess. I've been running a slightly updated version of bundler (due to needing my work laptop to remain useful for work). I'll make a seperate MR for that upgrade.

PR #195, opened 2023-12-20
Add config.secret_key_base

tom93:
secret_key_base was added in Rails 4 and replaces secret_token.

~~This PR is intended to merge using the "Rebase" strategy.~~ *Update: I reverted the second commit (see below), so now should use squash & merge and clean up the commit message.*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /config/unicorn.yml
 /config/isolate.yml
 /config/redis.yml
+/config/secret_key_base.*.txt
 /script/install.cfg
 /script/update.time
 /db/backup/
--- a/config/initializers/secret_key_base.rb
+++ b/config/initializers/secret_key_base.rb
@@ -1,18 +1,5 @@
 # Be sure to restart your server when you modify this file.
 
-# Your secret key for verifying the integrity of signed cookies.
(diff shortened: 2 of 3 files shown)


bagedevimo:
tested this fairly throughly in the `file` and `env` modes - pretty big fan, works lovely! i haven't tested the db version, but i can't see any reason it wouldn't work in review.

tom93:
Thanks!

bagedevimo (approved):

Holmes98 (dismissed):
Note that this invalidates Devise [remember tokens](https://www.rubydoc.info/github/heartcombo/devise/main/Devise/Models/Rememberable), ~~so we should deploy at the same time as #162 (which invalidates sessions)~~ on second thought, I don't think this matters (but it does appear to invalidate remember tokens).

tom93:
Oops. As per the chat, I made a mistake -- we were already using signed-but-not-encrypted cookies, just not for the session. So removing secret_token breaks e.g. the "remember me" feature. I reverted that commit, to fix the issue and allow automatic migration. We should keep secret_token for a while to give users a chance to open the site and get migrated.

So this should be good to merge now.

Holmes98 (approved):

PR #199, opened 2023-12-21
Update to Rails 4.2

bagedevimo:
Steps:
* Update Gemfile
* bundle update --conservative rails actionpack railties activerecord activemodel activesupport rack responders

There are lots of warnings & deprecations, but nothing is _failing_ in
the test suite.


bagedevimo:
Current dependencies on/for this PR:
* `master`
  * **PR #199** <a href="https://app.graphite.dev/github/pr/NZOI/nztrain/199?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #200** <a href="https://app.graphite.dev/github/pr/NZOI/nztrain/200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/NZOI/nztrain/199?utm_source=stack-comment).

config/secrets.yml
```diff
@@ -0,0 +1,22 @@
...
+
+# Make sure the secret is at least 30 characters and all random,
+# no regular words or you'll be exposed to dictionary attacks.
+# You can use `rake secret` to generate a secure secret key.
+
+# Make sure the secrets in this file are kept private
+# if you're sharing your code publicly.
+
+development:
+  secret_key_base: 0d886d0d879ad0bc8a0741aa6cf6ae7ed832c298a5e4f647bcfe147d7ee1e238ab0e0d519c96ff1890d7c8fc681765b4f9a183d039d0ff1314c21ae6b682de8f
```
tom93, line 14: We don't need this file (thanks to #195).

Holmes98 (dismissed):

tom93 (commented):
There is a [change to Rack](https://github.com/rack/rack/commit/28b014484a8ac0bbb388e7eaeeef159598ec64fc) that makes it listen on localhost instead of 0.0.0.0 in development to only allow connections from the current host ([source](https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-4-1-to-4-2.html)).

A side effect of this is that on some systems localhost might resolve to IPv6, which means connecting via 127.0.0.1 fails. I'd like to be able to continue using 127.0.0.1; looking into this.

Edit: [Better (official) source](https://guides.rubyonrails.org/4_2_release_notes.html#default-host-for-rails-server>). It says "both http://127.0.0.1:3000/ and http://localhost:3000/ will continue to work" but that's not the case for me. I know I can use `rails server -b 127.0.0.1` but that's annoying. I'm thinking of using the trick from [here](https://web.archive.org/web/20200616055654/https://fullstacknotes.com/make-rails-4-2-listens-to-all-interfaces/) (found via [Stack Overflow](https://stackoverflow.com/questions/28668436/how-to-change-the-default-binding-ip-of-rails-4-2-development-server/33249657#33249657) and the FastRuby article), but with 127.0.0.1 instead of 0.0.0.0. (Separate PR.)

Gemfile
```diff
@@ -3,7 +3,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
```
tom93, line 3: I'd like to add an explicit dependency on the 'responders' gem. From the ([release notes](https://guides.rubyonrails.org/4_2_release_notes.html#respond-with-class-level-respond-to)):
> `respond_with` and the corresponding class-level `respond_to` have been moved to the [responders](https://github.com/plataformatec/responders) gem
We use respond_with, but are saved because devise happens to depend on responders. Better to have explicit dependency.

tom93 (changes requested):

tom93:
Re commit message: I fixed two classes of deprecations in #219 and #220, I think there is only one left:

> DEPRECATION WARNING: The configuration option `config.serve_static_assets` has been renamed to `config.serve_static_files` to clarify its role (it merely enables serving everything in the `public` folder and is unrelated to the asset pipeline). The `serve_static_assets` alias will be removed in Rails 5.0. Please migrate your configuration files accordingly. (called from block in <top (required)> at /home/user/nztrain/config/environments/test.rb:37)

bagedevimo:
> Minor:
> 
> There is a [change to Rack](https://github.com/rack/rack/commit/28b014484a8ac0bbb388e7eaeeef159598ec64fc) that makes it listen on localhost instead of 0.0.0.0 in development to only allow connections from the current host ([source](https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-4-1-to-4-2.html)).
> 
> A side effect of this is that on some systems localhost might resolve to IPv6, which means connecting via 127.0.0.1 fails. I'd like to be able to continue using 127.0.0.1; looking into this.
> 
> Edit: [Better (official) source](https://guides.rubyonrails.org/4_2_release_notes.html#default-host-for-rails-server%3E). It says "both http://127.0.0.1:3000/ and http://localhost:3000/ will continue to work" but that's not the case for me. I know I can use `rails server -b 127.0.0.1` but that's annoying. I'm thinking of using the trick from [here](https://web.archive.org/web/20200616055654/https://fullstacknotes.com/make-rails-4-2-listens-to-all-interfaces/) (found via [Stack Overflow](https://stackoverflow.com/questions/28668436/how-to-change-the-default-binding-ip-of-rails-4-2-development-server/33249657#33249657) and the FastRuby article), but with 127.0.0.1 instead of 0.0.0.0. (Separate PR.)

Ha, I didn't notice this because i've been doing local development with `-b 0.0.0.0` intentionally

bagedevimo:
> Re commit message: I have draft fixes for two classes of deprecations in #219 and #220, only one left:
> 
> > DEPRECATION WARNING: The configuration option `config.serve_static_assets` has been renamed to `config.serve_static_files` to clarify its role (it merely enables serving everything in the `public` folder and is unrelated to the asset pipeline). The `serve_static_assets` alias will be removed in Rails 5.0. Please migrate your configuration files accordingly. (called from block in <top (required)> at /home/user/nztrain/config/environments/test.rb:37)

Awesome - I still think it's worth merging these all separately and potentially even deploying, then merging the fixes. Makes it a little easier to spot issues that occurred due to the upgrade vs due to fixing the deprecs - but i'm not very passionate about this.

Gemfile
Holmes98: Sounds good to me. Feel free to just commit here; I'm sure @bagedevimo is fine with it.

Holmes98 (approved):

tom93 (approved):

PR #204, opened 2023-12-22
Remove our String.shorten extension

tom93:
Rails 3 has a built-in truncate method that we can use instead.

Also remove the constant SHORTEN_LIMIT, it is only used in two places and removing it makes it easier to update the configuration.

Suggested-by: \@bagedevimo

~~Review and merge after #203 because of merge conflicts.~~ *Done*

*Edit:*

The place that uses it is actually dead code. Updating it anyway, can be deleted separately (draft: #205).

### DIFF
--- a/app/views/test_cases/index.html.erb
+++ b/app/views/test_cases/index.html.erb
@@ -18,8 +18,8 @@
 
 <% @test_cases.each do |test_case| %>
   <tr>
-    <td><%= simple_format(test_case.input.shorten) %></td>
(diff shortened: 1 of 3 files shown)


Holmes98 (dismissed):
Note that `truncate(100)` returns a string of at most 100 characters, while `shorten` could return up to 103 characters. But this code is unused anyway, so it doesn't matter.

Holmes98 (approved):

PR #216, opened 2023-12-23
Restrict permissions on config/secret_key_base.*.txt

tom93:
I should have done this from the start in #195.

(Review notes: The noise is because I renamed the variable "file" to "filename", to distinguish from the File object.)

For deploying, this will have to be applied manually (using chmod).


Holmes98 (approved):

tom93:
Deployed and applied manually.

PR #222, opened 2023-12-28
Add an informational comment on why we've got mini_racer in the Gemfile

bagedevimo:


Gemfile
```diff
@@ -97,6 +97,12 @@ gem 'sass'
 gem 'sass-rails'
 gem 'coffee-rails'
 gem 'uglifier', '>=1.0.3'
+# Provide a JS runtime to execjs without needing to
+# have node, bin, or similar installed on the relevant
```
tom93, line 101: Should "bin" be "bun"?

bagedevimo: hilariously, no: https://bun.sh/

tom93: So what's "bin"?

bagedevimo: oh, shit, i misread your comment, and my comment. fml.

PR #224, opened 2023-12-28
Use the shopify fork of qless

bagedevimo:
The mainstream release of qless is very unmaintained, and doesn't work with modern redis versions at all.

Shopify forked, and added a series of batches which fix a few bugs and improve a few UI quirks. Their fork is now also unmaintained, but it's more recently unmaintained, so that's nice.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,8 +52,8 @@ gem 'pg'
 gem 'backup'
 
 # Redis and Background Processing
-gem 'redis', '< 4.0'
-gem 'qless'#, :github => 'ronalchn/qless', :branch => 'nztrain'
+gem 'redis'
+gem 'qless', :github => 'Shopify/qless'
 gem 'connection_pool'
 gem 'sinatra'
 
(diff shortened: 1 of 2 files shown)


tom93:
I think this is the right choice, but just pointing out that they have diverged -- there are also commits in upstream that are not in the Shopify fork, not sure if any are important.

Suggested commit message:

```
Switch to the Shopify fork of qless (#224)
```

```
The original release of qless is very unmaintained, and doesn't work
with redis >= 6.2.7 [1].

Shopify forked and added a series of patches which fix a few bugs and
improve a few UI quirks. Their fork is now also unmaintained, but it's
more recently unmaintained, so that's nice.

[1] https://www.github.com/seomoz/qless-core/issues/89
```

bagedevimo:
force pushed and included your commit message :)

tom93 (approved):
I had a look and didn't spot anything important that's only in the original repository.
Should I merge & deploy?

Holmes98:
Sure.

tom93:
Deployed.

Aside:
There is a minor issue in the Qless web interface (this issue already existed in the previous version).
On /qless/server/failed, there is a list at the top of the different error types (restyled in this version), and clicking them is supposed to jump to the failed jobs with that error type, but it doesn't do anything.
On the other hand, right click > Open Link in New Tab (or middle click) still works.
I don't think it's worth fixing.

PR #225, opened 2023-12-28
Commit updated db/schema.rb

bagedevimo:
By updating the rails version, we now get a slightly different dump format. In practise, this is the same effective result, but we should commit this new version to:
(a) ensure everyone is definitely loading the right thing when using
    schema:load
(b) prevent diff's when doing a deploy which is just scary looking for
    now reason.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -16,7 +16,7 @@ ActiveRecord::Schema.define(version: 20200418113601) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
-  create_table "contest_relations", force: true do |t|
+  create_table "contest_relations", force: :cascade do |t|
     t.integer  "user_id"
     t.integer  "contest_id"
     t.datetime "started_at"
@@ -40,7 +40,7 @@ ActiveRecord::Schema.define(version: 20200418113601) do
   add_index "contest_relations", ["contest_id", "user_id"], name: "index_contest_relations_on_contest_id_and_user_id", unique: true, using: :btree
   add_index "contest_relations", ["user_id", "started_at"], name: "index_contest_relations_on_user_id_and_started_at", using: :btree
 
-  create_table "contest_scores", force: true do |t|
+  create_table "contest_scores", force: :cascade do |t|
     t.integer  "contest_relation_id", null: false
     t.integer  "problem_id",          null: false
     t.integer  "score"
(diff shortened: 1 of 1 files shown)


tom93 (approved):
Typo in commit message: "now reason" => "no reason"
I think we also need to edit some of the migrations (separate PR), ~~otherwise a clean install using db:migrate produces more diff's to do with string length limits.~~
*Update 1:* I can't reproduce this any more.
*Update 2:* Apparently it's the other way round -- if you have an existing DB created using < 4.2 and you run db:schema:dump using Rails 4.2, it will add `limit: 255` to preserve semantics because new default is to not limit values: https://www.github.com/rails/rails/issues/19001. So I guess we can either explicitly add those limits in the migrations and schema, or update the production DB to remove those limits.

tom93 (changes requested):
(See my previous comment.)

Holmes98:
We discussed this earlier but another option is to [remove old migrations](https://stackoverflow.com/questions/4248682/is-it-a-good-idea-to-purge-old-rails-migration-files/26351147#26351147) (I believe this means we'll no longer inherit changes to the rails defaults; not sure if we care). I think my preference is to either do that, or just manually remove the limits from the production DB.

bagedevimo:
I'm in camp update the old migrations, because I don't like modifying the DB outside of migrations and a migration to remove the limits would be weird, if it didn't follow a migration that had the limits.

tom93:
How about inserting `limit: 255` into the existing migrations (to retain previous behaviour), and also adding a new migration to remove the limit (I think by changing the column type)?
Not sure if that's the best of both or the worst of both ;)

bagedevimo:
Yeah I'll do that. I think it's best option for git history and non-using-psqls sake

tom93 (approved):
Let's handle the `limit: 255` in a separate PR(s)?

Suggested commit message:

```
Commit updated db/schema.rb (#225)
```

```
Rails 4.2 introduced `force: :cascade` and uses that by default when
dumping the schema[1] to make it possible to reload a schema when
foreign keys are in place, so we now get a slightly different dump
format. There are also other trivial differences.

Commit this new version to reduce diff's when the schema is dumped.

Note that there are additional differences when dumping databases
created pre-4.2 (to do with `limit: 255`, see #225 for discussion),
but they will be addressed separately.

[1] https://guides.rubyonrails.org/v4.2/4_2_release_notes.html#active-record-notable-changes
```

PR #228, opened 2023-12-29
Setup capistrano for deployments

bagedevimo:
The current deployment process doesn't give us many options for
rollbacks, and could have issues if the server was restarted during the
deployment. Most rails apps that aren't containerised tend to be
deployed with Capistrano [1].

Capistrano lets us all the same user, and manage deployment access by
giving people access to the deploy user by adding their ssh public key.
We should be able to move towards that being a sudo-less enterprise, but
qless may make that difficult.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,0 +1,39 @@
+# config valid for current version and patch releases of Capistrano
+lock "~> 3.18.0"
+
+set :application, "nztrain"
+set :repo_url, "git@github.com:NZOI/nztrain.git"
+
+# Default branch is :master
+# ask :branch, `git rev-parse --abbrev-ref HEAD`.chomp
+
+# Default deploy_to directory is /var/www/my_app_name
+# set :deploy_to, "/var/www/my_app_name"
+
+# Default value for :format is :airbrussh.
+# set :format, :airbrussh
+
+# You can configure the Airbrussh format using :format_options.
+# These are the defaults.
+# set :format_options, command_output: true, log_file: "log/capistrano.log", color: :auto, truncate: :auto
+
+# Default value for :pty is false
+# set :pty, true
+
+# Default value for :linked_files is []
+append :linked_files, "config/database.yml", "config/redis.yml"
+
+# Default value for linked_dirs is []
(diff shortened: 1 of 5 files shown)


bagedevimo:
Current dependencies on/for this PR:
* `master`
  * **PR #228** <a href="https://app.graphite.dev/github/pr/NZOI/nztrain/228?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/NZOI/nztrain/228?utm_source=stack-comment).

config/deploy/production.rb
```diff
@@ -0,0 +1,57 @@
+# server-based syntax
+# ======================
+# Defines a single server with a list of roles and multiple properties.
+# You can define all roles on a single server, or split them:
+
+# server "train.nzoi.org.nz", user: "cap_deploy", roles: %w[app db web]
+server "faketrain", user: "cap_deploy", roles: %w[app db web]
```
tom93, line 7: Is this just from debugging?

bagedevimo: yeah, this needs some decisions - we could use a new user, or make some changes to the existing deploy user, and i wanted to make sure that if this got accidentally merged it couldn't do any damage.

config/deploy.rb
```diff
@@ -0,0 +1,42 @@
...
+
+# You can configure the Airbrussh format using :format_options.
+# These are the defaults.
+# set :format_options, command_output: true, log_file: "log/capistrano.log", color: :auto, truncate: :auto
+
+# Default value for :pty is false
+# set :pty, true
+
+# Default value for :linked_files is []
+append :linked_files, "config/database.yml", "config/redis.yml"
```
Holmes98, line 24: I think we'll also need `config/isolate.yml`, `config/secret_key_base.production.txt`, `script/install.cfg` to persist between releases. There's also a few other config files listed in .gitignore that could be worth adding, though I think we don't currently use them.

Also I think assets under `db/data` should go under linked_dirs?

bagedevimo: I have no idea what's going on with db/data is is spooking me out - added the rest though.

Holmes98 (approved):

PR #230, opened 2023-12-29
Make /users etc. twice as fast by preloading roles

tom93:
From the commit message:

> This is an example of the N+1 query problem. Fixing it makes the pages render about twice as fast when there are many (thousands) of users.
> 
> The rest of the time is spent in the view, which is harder to optimise (there are some low-hanging fruit such as caching the policies, but they'll only get us so far; the only complete solution is to introduce paging).
> 
> See: #192
> 
> Reported-by: @thebrucecgit (and others)

For testing, the following snippet can be used to populate the database with 10,000 users:

```ruby
def create_user(id)
  attrs = {username: "u#{id}", email: "u#{id}@example.com", name: "U#{id}", confirmed_at: Time.now.utc}
  User.new(attrs).save!(validate: false)
end

User.transaction { (User.count + 1).upto(10000) { |i| create_user(i) } }
```

I suggest running it using `rails runner '...'` rather than `rails console`, so it won't spam the terminal with log messages.

---

If anyone has ideas on how to make the view even faster (without resorting to paging), please chime in! The relevant code is in [app/views/users/index.html.erb](https://github.com/NZOI/nztrain/blob/5a712db9966350b7046b8523d0b581f508e270c5/app/views/users/index.html.erb#L31:L55). I have a draft patch that caches the policies in branch [users-index-faster-cache-policies](https://github.com/NZOI/nztrain/commit/ed8c68f435c6708331d840fad8adc0fa6cf7f766), but it doesn't make that much difference. I suspect we'll have to switch to paging.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,19 +1,19 @@
 class UsersController < ApplicationController
   def index
     authorize User.new, :index?
-    @users = policy_scope(User).num_solved.order(:email)
+    @users = policy_scope(User).num_solved.includes(:roles).order(:email)
     render
   end
 
   def online
     authorize User.new, :inspect?
-    @users = policy_scope(User).order(:last_seen_at).reverse_order
+    @users = policy_scope(User).includes(:roles).order(:last_seen_at).reverse_order
     render
   end
 
   def newest
     authorize User.new, :inspect?
-    @users = policy_scope(User).order(:created_at).reverse_order.limit(100)
+    @users = policy_scope(User).includes(:roles).order(:created_at).reverse_order.limit(100)
     render
   end
 


Holmes98 (approved):
Looks good. I ran some benchmarks (below) for `/users` on a copy of our prod database; didn't quite get a 2x speedup but it's significantly better. Agree that policy caching doesn't make much difference; I don't think it's worth adding.

master (5a712db9966350b7046b8523d0b581f508e270c5):
```
Completed 200 OK in 3543ms (Views: 3099.1ms | ActiveRecord: 400.7ms)
Completed 200 OK in 3300ms (Views: 2885.3ms | ActiveRecord: 405.1ms)
Completed 200 OK in 3455ms (Views: 3013.9ms | ActiveRecord: 424.6ms)
Completed 200 OK in 3471ms (Views: 3034.9ms | ActiveRecord: 426.5ms)
Completed 200 OK in 3530ms (Views: 3092.2ms | ActiveRecord: 428.7ms)
```

users-index-faster (d9bcd562bd72be619cc3a3c2d5b0d31a1f3155e2)
```
Completed 200 OK in 2827ms (Views: 2729.7ms | ActiveRecord: 55.1ms)
Completed 200 OK in 2346ms (Views: 2300.6ms | ActiveRecord: 38.0ms)
Completed 200 OK in 2569ms (Views: 2514.2ms | ActiveRecord: 46.0ms)
Completed 200 OK in 2485ms (Views: 2438.5ms | ActiveRecord: 37.7ms)
Completed 200 OK in 2706ms (Views: 2655.2ms | ActiveRecord: 42.1ms)
```

users-index-faster-cache-policies (ed8c68f435c6708331d840fad8adc0fa6cf7f766)
```
Completed 200 OK in 2672ms (Views: 2580.1ms | ActiveRecord: 52.7ms)
Completed 200 OK in 2230ms (Views: 2179.7ms | ActiveRecord: 41.6ms)
Completed 200 OK in 2439ms (Views: 2381.2ms | ActiveRecord: 37.8ms)
Completed 200 OK in 2536ms (Views: 2488.4ms | ActiveRecord: 37.4ms)
Completed 200 OK in 2346ms (Views: 2295.5ms | ActiveRecord: 41.4ms)
```

I'd prefer not to paginate, or at least make it optional (or have an adjustable page size), since it's much easier to search for something when everything's on the same page.

For scripting purposes we could potentially add a separate endpoint to look up a user by username. We could also add XML/JSON endpoints for `/users` (json seems faster, renders in ~1500ms for me and could possibly be improved by excluding fields). But for https://github.com/NZOI/nztrain/issues/192#issuecomment-1651658313 it's probably best to just integrate NZIC registration into nztrain itself.

tom93:
Thanks for benchmarking, I'll tone down my claimed speed improvement :)
Might be to do with the number of users, investigating...

tom93:
Here are the timings in prod, definitely improved but not as much as I originally expected.

Before:
```
Completed 200 OK in 5223ms (Views: 4383.0ms | ActiveRecord: 831.3ms)
Completed 200 OK in 5255ms (Views: 4437.3ms | ActiveRecord: 805.8ms)
Completed 200 OK in 5162ms (Views: 4340.2ms | ActiveRecord: 812.4ms)
Completed 200 OK in 5192ms (Views: 4348.5ms | ActiveRecord: 834.2ms)
```
After:
```
Completed 200 OK in 4196ms (Views: 4051.3ms | ActiveRecord: 97.6ms)
Completed 200 OK in 3934ms (Views: 3812.8ms | ActiveRecord: 110.1ms)
Completed 200 OK in 3832ms (Views: 3741.0ms | ActiveRecord: 80.5ms)
Completed 200 OK in 4109ms (Views: 4016.2ms | ActiveRecord: 82.1ms)
```

PR #176, opened 2023-02-19
Update to latest version of isolate

Holmes98:
Also see https://github.com/NZOI/isolock/commit/94b13d35eecca18665d1dbd919ed718dfb078e1d for required changes to isolock.

PR #235, opened 2024-01-01
Prevent other users from executing 'isolate'

tom93:
When we updated to latest version of isolate (#176), the permissions changed from 4750 to 4755 (the upstream default, allowing all users to read/execute).

On further thought, it's probably safer to prevent other users from executing isolate, just because it's setuid.

### DIFF
--- a/script/install/isolate.bash
+++ b/script/install/isolate.bash
@@ -25,3 +25,6 @@ make install || {
   rm -r isolate
   exit 1
 }
+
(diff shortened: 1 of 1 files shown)


tom93:
Technically there is a race condition between `make install` and the `chmod`, but I don't think there is an easy way to fix it and it's unlikely to cause harm.

Holmes98 (approved):

tom93:
(Deployed manually.)

PR #223, opened 2023-12-28
Add GitHub Actions workflow for CI

Holmes98:
This is an alternative to #82 that does not use a Dockerfile.

Note that the commands used are mostly copied from our previous CI setup in [.travis.yml](https://github.com/NZOI/nztrain/blob/edd8b4bae70ff445623e0448d1eda690940f0281/.travis.yml) (which is no longer being used due to #147).

PR #236, opened 2024-01-01
Use correct group name for isolate/isolock

tom93:
Previously the scripts assumed that the group name is the same as the username, but that's not necessarily the case.

This came up in #223.

Merge using "rebase".

### DIFF
--- a/script/install.cfg.template
+++ b/script/install.cfg.template
@@ -14,6 +14,7 @@ export APP_NAME=${APP_NAME}
 # Rails environment that will be used to run this installation
 export RAILS_ROOT=${RAILS_ROOT}
 export APP_USER=${APP_USER} # user this app will run under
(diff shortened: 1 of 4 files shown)


Holmes98 (approved):

tom93:
Deployed new permissions of isolock manually.
Manually added APP_GROUP to script/install.cfg.
While doing that, I noticed APP_USER was set to "nztrain", which is odd because the web server currently runs as the "deploy" user (I couldn't find where we set that user, so I assume it's due to [auto detection](https://www.phusionpassenger.com/library/indepth/meteor/app_autodetection/nginx/)). I changed it to "deploy" (though it doesn't really matter because APP_USER is basically unused).
I'll open an issue to decide if we want to make the app run as the "nztrain" user.

PR #237, opened 2024-01-01
Add env var DATABASE_INSTALL to skip PostgreSQL install

tom93:
And a lot of cleanup. Came up from #223.

(The GitHub runner already has the postgresql package installed, so this won't save an install in the current setup, but it may save an install in other environments.)

Comments:
I don't know what config/database.yml.default is for (has been there since first commit, appears unused). We should eventually get rid of the install process that generates database.yml from the template, and at that point there would just be a database.yml that works out-of-the-box -- no need for a sample.

Merge using "rebase".


Holmes98 (approved):
Works with the default configuration. Doesn't work with a non-default DATABASE_USERNAME (using peer authentication), but that didn't work before anyway, it just fails earlier now.

tom93:
(Deployed the trivial change to config/database.yml manually.)

PR #240, opened 2024-01-05
Update config.serve_static_assets to config.serve_static_files

bagedevimo:
This option is deprecated in Rails 4.2 and removed in Rails 5. There isn't a lot of rational explained in the MR, but hey.

[1] https://github.com/rails/rails/pull/18100
[2] https://github.com/rails/rails/commit/463b5d7581ee16bfaddf34ca349b7d1b5878097c

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -33,7 +33,7 @@ NZTrain::Application.configure do
 
   # Disable Rails's static asset server
   # In production, Apache or nginx will already do this
-  config.serve_static_file = true # true so Webrick testing of production mode works
+  config.serve_static_files = true # true so Webrick testing of production mode works
 
   # Enable serving of images, stylesheets, and javascripts from an asset server
   # config.action_controller.asset_host = "http://assets.example.com"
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -33,7 +33,7 @@
 
   # Disable Rails's static asset server
   # In production, Apache or nginx will already do this
-  config.serve_static_assets = true # true so Webrick testing of production mode works
+  config.serve_static_file = true # true so Webrick testing of production mode works
```
tom93, line 36:
```suggestion
  config.serve_static_files = true # true so Webrick testing of production mode works
```

tom93 (changes requested):

tom93 (approved):
Was unclear to me at first, but it looks like a simple rename (only the old alias will be removed, the configuration option will continue to exist under the new name).

Suggested commit message to clarify this:

```
Update config.serve_static_assets to config.serve_static_files
```

```
The configuration option `config.serve_static_assets` has been renamed
to `config.serve_static_files in Rails 4.2 [1], and the old name will
be removed in Rails 5.0 [2].

[1] https://www.github.com/rails/rails/pull/18100
[2] https://github.com/rails/rails/commit/463b5d7581ee16bfaddf34ca349b7d1b5878097c
```

PR #241, opened 2024-01-05
Send deprecation notices to the rails log

bagedevimo:
Prior to this change, messages were being send to the ActiveSupport::Notifications system which doesn't appear to have any active listeners, meaning these notices were just being sent into the void. The notification was probably setup for a old integration (maybe newrelic?) but nothing appears to be using it at the moment.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -49,7 +49,7 @@ NZTrain::Application.configure do
   config.i18n.fallbacks = true
 
   # Send deprecation notices to registered listeners
-  config.active_support.deprecation = :notify
+  config.active_support.deprecation = :log
 
   # Compress JavaScript and CSS  
   config.assets.compress = true  


tom93:
The [docs](https://api.rubyonrails.org/v4.2.0/classes/ActiveSupport/Notifications.html#method-c-subscribe) say

> Notifications ships with a queue implementation that consumes and publishes events to all log subscribers.

I'm curious why that doesn't seem to happen for us; would like to investigate.

bagedevimo:
I'm not sure it means that version of "log" - here's that implementation it describes https://github.com/rails/rails/blob/30af171af13293aa37bee612ae7b976d3ede0439/activesupport/lib/active_support/notifications/fanout.rb - i think it's meaning more like in sense of a message queue log ala kafka

bagedevimo:
> (Aside: deprecation = :notify is the default.)
TIL, wild.

tom93 (approved):
(Didn't test)

> i think it's meaning more like in sense of a message queue log

Aha.

Rails 7 [goes further](https://www.github.com/rails/rails/pull/42913) and completely disables deprecation warnings in production by default, so I suppose it's good to explicitly set this to a non-default value. But it might also be interesting to subscribe to notifications and see if we're missing out on anything useful.

Suggested commit message:

```
Send deprecation notices to the rails log
```

```
Prior to this change, messages were being sent to the
ActiveSupport::Notifications system in production (as per the Rails
default) which doesn't appear to have any active listeners, meaning
these notices were just being sent into the void.

Send the messages to the log instead, so that we can see them.
```

PR #242, opened 2024-01-05
Update devise

bagedevimo:
Required for Rails 5

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ ruby '>= 2.4.10', '< 2.5'
 gem 'rails', '~> 4.2'
 gem 'json_cve_2020_10663', '~> 1.0' # required until we update json >= 2.3, which we can only do once we upgrade to Rails >= 4.2 because activesupport 4.1.* depends on json ~> 1.7 (i.e < 2.0): https://rubygems.org/gems/activesupport/versions/4.1.16
 
-gem 'devise', '~> 3.4.1'
+gem 'devise', '~> 4.0'
 gem 'psych', '~> 2.0.2' # part of stdlib, need newer version for safe_load
 
 gem 'rubyzip', '1.3.0'
--- a/Gemfile
+++ b/Gemfile
@@ -86,6 +86,8 @@ group :development, :test do
   gem 'byebug'
 
   gem 'standard'
+
+  gem "test_after_commit", "~> 1.2"
 end
 
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -109,12 +109,11 @@ GEM
     crass (1.0.5)
     daemons (1.4.1)
     debug_inspector (0.0.3)
-    devise (3.4.1)
+    devise (4.9.3)
       bcrypt (~> 3.0)
       orm_adapter (~> 0.1)
-      railties (>= 3.2.6, < 5)
+      railties (>= 4.1.0)
       responders
-      thread_safe (~> 0.1)
       warden (~> 1.2.3)
     diff-lcs (1.5.0)
     docile (1.3.5)
@@ -442,7 +441,7 @@ DEPENDENCIES
   connection_pool
   countries
   country_select
-  devise (~> 3.4.1)
+  devise (~> 4.0)
   facebox-rails
   factory_bot_rails
   foreman
@@ -500,4 +499,4 @@ RUBY VERSION
    ruby 2.4.10p364
 
 BUNDLED WITH
-   1.16.1
+   1.17.3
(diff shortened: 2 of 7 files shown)


spec/controllers/accounts/registrations_controller_spec.rb
```diff
@@ -1,30 +1,36 @@
 require 'spec_helper'
 
 describe Accounts::RegistrationsController do
-  it "can get signup form" do
+  it 'can get signup form' do
```
Holmes98, line 4: Is there a particular reason for changing to single quotes? standardrb seems to prefer double quotes:
> Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping

spec/controllers/accounts/registrations_controller_spec.rb
```diff
@@ -1,30 +1,36 @@
...
     get :new
     expect(response).to be_success
   end
 
   it 'can signup (create action)' do
     expect do
-      post :create, :user => { :username => "signup_username", :name => "Mr. SignUp", :email => "signup@nztrain.com", :password => "password", :password_confirmation => "password" }
-    end.to change{User.count}.by(1)
+      post :create,
+           user: { username: 'signup_username', name: 'Mr. SignUp', email: 'signup@nztrain.com', password: 'password',
```
Holmes98, line 12: This also fails lint:
> Layout/ArgumentAlignment: Use one level of indentation for arguments following the first line of a multi-line method call

Holmes98 (dismissed):
Seems fine to me, other than the lint issues. I have slight preference to using [test_after_commit](https://github.com/heartcombo/devise/blob/main/CHANGELOG.md#410) over commenting out the specs, but I don't feel particularly strongly about it.

spec/controllers/accounts/registrations_controller_spec.rb
bagedevimo: fixed now! only 2 years later. #nailed it.

spec/controllers/accounts/registrations_controller_spec.rb
```diff
@@ -1,30 +1,36 @@
...
     # check signup attributes saved
     newuser = User.find_by_username("signup_username")
     expect(newuser).not_to be_nil
     expect(newuser.name).to eq("Mr. SignUp")
     expect(newuser.email).to eq("signup@nztrain.com")
     expect(newuser.valid_password?("password")).to be true
+
+    # Due to how transactions are used in tests under Rails < 5
+    # these tests don't work on modern devise (fixed / broken in 4.1.0
+    # TODO: Re-enable these lines after we're on rails 5
```
Holmes98, line 24: This can be removed now (in which case the rest of the changes to this file are formatting changes only).

Sidenote: I'm kinda keen to just run `standardrb --fix` over the whole codebase and commit those changes in a separate PR, so we aren't mixing formatting changes with actual changes (otherwise it's a bit harder to review). Any thoughts?

Gemfile
```diff
@@ -105,3 +105,5 @@ gem 'uglifier', '>=1.0.3'
 gem 'mini_racer', '~> 0.4.0'
 gem 'yui-compressor'
 #end
+
+gem "test_after_commit", "~> 1.2"
```
Holmes98, line 109: Should we move this to the development/test group?

Holmes98 (commented):
Commit history is also a bit messy with 78d3c76 introducing conflict markers and not actually commenting out the tests. I don't mind if we're going to squash this PR, otherwise I think we can just drop that commit in a rebase?

spec/controllers/accounts/registrations_controller_spec.rb
bagedevimo: keen & happy to do, but can we merge these couple of MRs first in the interest of loooong running?

Gemfile
bagedevimo: Done!

bagedevimo:
> Commit history is also a bit messy with [78d3c76](https://github.com/NZOI/nztrain/commit/78d3c769143f617c35bde549d516ff9ee4e78eec) introducing conflict markers and not actually commenting out the tests. I don't mind if we're going to squash this PR, otherwise I think we can just drop that commit in a rebase?

I had the same thought and already done so :)

Holmes98:
I'm just going to amend that last commit message to remove this (and fix a couple typos)
> Restore missing tests and run standardrb over a changed file to bring
the formatting into compliance with the declared config

Holmes98 (approved):

PR #250, opened 2024-01-05
Restore :assets group in Gemfile

tom93:
**TODO:** https://github.com/NZOI/nztrain/pull/250#issuecomment-1879579742

Some of those gems (especially mini_racer) are very large, and it is useful to be able to skip them. They were already in a commented-out group; restore the group to allow skipping them (they will still be installed by default).

Note that putting them in a group will mean they won't be auto-required when config/application.rb calls
`Bundler.require(:default, Rails.env)`, but that should be fine.

Also remove the old comment "not required in production environments by default" because it is unclear/inaccurate (we currently run the asset precompilation on the production server), and add hints on how to skip the group.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -89,19 +89,19 @@ group :development, :test do
 end
 
 
-# Gems used only for assets and not required  
-# in production environments by default.  
-#group :assets do  
-gem 'sass'
-gem 'sass-rails'
-gem 'coffee-rails'
-gem 'uglifier', '>=1.0.3'
-# Provide a JS runtime to execjs without needing to
-# have node, bun, or similar installed on the relevant
-# server. We should write this out asap, along with the
-# rest of the gems in this assets category, as bundling
-# a version of libv8 into a rubygem is just security-vuln
-# city, but hey, what can you do.
-gem 'mini_racer', '~> 0.4.0'
-gem 'yui-compressor'
-#end
+# Gems used only for precompiling assets. They may be skipped using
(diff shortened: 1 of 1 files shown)


bagedevimo (approved):

Holmes98 (dismissed):
I'm not quite sure I fully understand why, but the assets group was removed in Rails 4:
- https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0-gemfile
- https://stackoverflow.com/questions/16406204/why-did-rails4-drop-support-for-assets-group-in-the-gemfile
- rails/rails@49c4af43ec5819d8f5c1a91f9b84296c927ce6e7

And this doesn't seem ideal, but we might be compiling some assets live in production mode; not sure if that could cause issues with this:
https://github.com/NZOI/nztrain/blob/b084e51ec878bb3ff0fbacbade4fbc456f97ba14/config/environments/production.rb#L59-L60

tom93:
Good catch, what a mess. Just wanted to skip some gems...
I'm tempted to keep the assets group (even though Rails got rid of it), because that seems like the way to allow those gems to be skipped.
I'll definitely first double check that asset precompilation still works identically (some of those comments about needing to patch the rake task and missing initializers got me worried).

> we might be compiling some assets live in production mode

I hope that can be dealt with separately? (The gems will still be installed in production, just won't be auto-required by default; I'll try to confirm that live compilation still works with that change.)

PR #254, opened 2024-01-06
Add system/mailer settings to db/seeds.rb

Holmes98:
> These are currently created during migrations, but we plan to remove this behaviour in #243. The values are left blank, as they need to be set manually anyway.

@tom93 suggested using an environment variable for the password; I didn't bother, but feel free to add that if you prefer.

Commit message can be changed if we decide not to go with #243.


bagedevimo (approved):

bagedevimo:
IMO this is worth doing even regardless of migration changes, as it makes the whole thing more complete with just a db:seed

Holmes98:
Done in #273.

PR #258, opened 2024-01-08
Fix crash when adding file attachment to problem

tom93:
Previously, if there was a validation error when adding a file attachment to a problem (e.g. "Filepath has already been taken", "Filepath extension doesn't match file"), then instead of displaying the validation error message, there would be a 500 Internal Server Error (First argument in form cannot contain nil or be empty). Full trace below.

The cause was an incomplete renaming in commit 1e931189 (enable file attachments for problems, with shared code for filelinks of groups and problems, 2013-12-19). The code path that triggers when there is a validation error still used some old variable names, causing the crash.

Fix by updating those variable names.

Full error trace:

```
Completed 500 Internal Server Error in 764ms (ActiveRecord: 64.1ms)
First argument in form cannot contain nil or be empty excluded from capture: DSN not set

ActionView::Template::Error (First argument in form cannot contain nil or be empty):
    1: <%= form_for @new_filelink, :url => index_path do |f| %>
    2:   <% if @new_filelink.errors.any? %>
    3:     <div id="error_explanation">
    4:       <h2><%= pluralize(@new_filelink.errors.count, "error") %> prohibited this post from being saved:</h2>
  app/views/filelinks/roots/index.html.erb:1:in `_app_views_filelinks_roots_index_html_erb___4252713757973532291_70289880'
  app/controllers/filelinks/roots_controller.rb:88:in `create'
```


Holmes98 (approved):
Note that it still ~~crashes~~ throws an error when entering an invalid file attachment ID, otherwise looks good.

tom93:
Thanks.
(When the file attachment ID is invalid I get a 404 page in prod env: "The page you were looking for doesn't exist." Should also be fixed, but I don't think it's as bad as a crash.)

PR #260, opened 2024-01-28
Add support for custom evaluators using compiled langauges

Holmes98:
This adds an optional language field to the Evaluator class. All existing/new evaluators will default to having no language, which maintains the existing functionality (script is executed as-is).

Note that C++ evaluators were already "supported" through [cint.rb](https://github.com/NZOI/nztrain/blob/8ffcd6a82ccef82d078df33863d4d6b41e1054a3/script/install/cint.rb), but that is much slower as it requires the evaluator to be recompiled on every test case. That functionality can be considered deprecated and removed later, after fixing existing evaluators.

Also note that for evaluators with no language, Pygments will attempt to automatically detect the language based on the code (see [here](https://github.com/pygments/pygments.rb/blob/7ef665b4a1247d367f5ba7a018bc0d2baf183852/lib/pygments/mentos.py#L93)).

### DIFF
--- a/app/workers/judge_submission_worker.rb
+++ b/app/workers/judge_submission_worker.rb
@@ -52,7 +52,7 @@ class JudgeSubmissionWorker < ApplicationWorker
     raise
   end
 
-  EvalFileName = "eval.sh"
+  EvalFileName = "eval"
   OutputBaseLimit = 1024 * 1024 * 2
 
   attr_accessor :submission, :exe_filename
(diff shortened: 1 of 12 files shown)


app/workers/judge_submission_worker.rb
```diff
@@ -68,14 +68,34 @@ def judge
...
+      else
+        if problem.evaluator.language.nil?
+          eval_command = "./#{EvalFileName}"
+        else
+          eval_command = problem.evaluator.language.run_command(EvalFileName)
+        end
+
+        if problem.evaluator.language&.compiled
+          evaluator_compilation = compile!(problem.evaluator.source, problem.evaluator.language, EvalFileName) # possible caching
+          raise evaluator_compilation['log'] if evaluator_compilation['stat'] != 0 # error
```
BelgianSalamander, line 90: It seems a little strange to raise an error if evaluator compilation fails. This leaves submissions pending and doesn't make it clear that evaluator compilation failed and why it failed (without looking on qless). Also, people with the staff role have the permission to create evaluators but wouldn't be able to see their evaluator's compilation errors.

I imagine it's sensible to set the status to the submission as errored (2) when this occurs instead of raising an error and then store + display the evaluator compilation log to staff/admin viewing the submission.

BelgianSalamander (commented):
I don't know too much ruby (or rails) and am not too familiar with the codebase but the code changes seem to make sense.

I tested it locally for a bit and it worked as expected but I'm not sure about raising an error when evaluator compilation fails.

app/workers/judge_submission_worker.rb
Holmes98: Thanks for reviewing. I think this is a reasonable suggestion, and have pushed some changes to address this.

BelgianSalamander:
This is probably outside the scope of this pr but in future it may be good to not display the message `This submission has not finished judging. Refresh this page in a minute or two to see the submission's score.` when judging encounters an error (be it evaluator not compiling or some other error in the worker).

Holmes98:
Yeah, the submission page doesn't really handle judging errors. I've committed some improvements in 9e7aa3e.

BelgianSalamander (approved):

Holmes98:
CI is currently failing; need to merge #266 first.

BelgianSalamander:
Seems to be fixed now

Holmes98:
Deployed and converted all existing C++ evaluators that were previously using cint.rb to use this instead.

PR #261, opened 2024-01-28
Add support for interactive problems

Holmes98:
Depends on #260. Based on the [Communication](https://cms.readthedocs.io/en/latest/Task%20types.html#communication) task type from CMS.

This adds an `interactive_processes` field to the Evaluator class.

If interactive_processes=0 (default), functionality is unchanged (evaluator runs after the user/submission program).

If interactive_processes>0, we concurrently spawn that many user processes, each connected using pipes to the grader process. After all processes exit, execution stats are merged (time/memory is added, wall time is maximised, first failing status is used).

* Unlike normal evaluators, interactive evaluators (which I'm referring to as "graders") must always terminate gracefully, otherwise it is judged as en evaluator error (see [here](https://www.github.com/cms-dev/cms/issues/993) for relevant discussion). Need to handle SIGPIPE when writing to the user pipes, in case they crashed/exited early, etc.
* There is currently some extra handling to make MLE/TLE report correctly after merging stats; it feels a bit hacky but it seemed like the easiest approach (happy to take suggestions though).
* If interactive_processes>1, the process number is sent as a command line argument to each user program, but it can also be sent by the grader if preferred.
* I'm using file descriptors 5+ for the user pipes, fd 4 for "expected output", and stdin for the case input. fd 3 is currently unused; it can be considered "reserved" in case we want to add more functionality later (e.g. for "actual output" that can be shown to the user for sample cases, or for staff to debug). I'm also considering swapping file descriptors stdin/3 for non-interactive evaluators (currently 3 => case input, stdin => actual output) so they match, but have left it for now (will require manually updating existing evaluators).
* `interactive_processes` is currently limited to 2 to avoid exhausting server resources.

### DIFF
--- a/app/controllers/evaluators_controller.rb
+++ b/app/controllers/evaluators_controller.rb
@@ -2,7 +2,7 @@ class EvaluatorsController < ApplicationController
 
   def permitted_params
     @_permitted_params ||= begin
-      permitted_attributes = [:name, :description, :source, :language_id]
+      permitted_attributes = [:name, :description, :source, :language_id, :interactive_processes]
       permitted_attributes << :owner_id if policy(@evaluator || Evaluator).transfer?
       params.require(:evaluator).permit(*permitted_attributes)
     end
--- a/app/models/evaluator.rb
+++ b/app/models/evaluator.rb
@@ -6,5 +6,6 @@ class Evaluator < ActiveRecord::Base
   belongs_to :language
 
   validates :name, :presence => true
+  validates :interactive_processes, :presence => true, :numericality => { :greater_than_or_equal_to => 0, :less_than_or_equal_to => 2 }
 
 end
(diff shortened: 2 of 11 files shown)


Holmes98:
I've marked this as a draft but am looking for feedback if anyone is able to review.

BelgianSalamander:
I have tested this on IOI 2018 Combo and IOI 2020 Stations. No glaring issues so far. I will try to review soon.

BelgianSalamander (commented):
This seems mostly correct overall. 

I like the idea of fd 3 being used for "actual output" as for some problems it definitely makes sense to show some extra feedback to users on sample cases (as currently the only feedback is the testcase outcome). 

While testing this with the IOI problems, I found it difficult to debug evaluators and submissions as it can usually be unclear if the problem lies in the evaluator or the submission. I feel like it could be useful to have some way of showing the stderr of the user processes to staff to make it easier when making problems as otherwise there is no way to get any info directly from the user processes.

PR #267, opened 2024-05-15
Improve judging errors on the submission page

Holmes98:
Depends on #260. This adds a proper error message and hides test cases on the submission page when judging fails.

Intending to squash-merge.


app/models/submission.rb
```diff
@@ -149,7 +149,7 @@ def update_test_messages
 
       judge_data = self.judge_data
       return if judge_data.status == :pending # incomplete judge_log
-      return if judge_data.errored? # judge errored - very bad
+      return if judge_data.data.has_key?('error') # judge errored - very bad
```
BelgianSalamander, line 152: Is it normal that no testing errors are given if the judge log has the error key? From my understanding this would only happen if a ruby error occurs while judging (notably when a problem has no test cases). I feel like this sort of error is also something we would want to be reported to problem authors.

BelgianSalamander (dismissed):

PR #268, opened 2024-05-19
Display an error message when submitting executable.

### LINKED ISSUE #257: Produce nice error message when .exe submitted instead of source code
Currently, if a user tries to submits and uploads their compiled .exe instead of their source file, it causes a 500 Internal Server Error (invalid byte sequence in UTF-8). We should instead display a helpful error message. (Note that this error could also happen if the source file has the wrong encoding.)

Log:

```
Completed 500 Internal Server Error in 112ms (ActiveRecord: 58.2ms)
invalid byte sequence in UTF-8 excluded from capture: DSN not set

ArgumentError (invalid byte sequence in UTF-8):
  app/controllers/problems_controller.rb:90:in `block in submit'
  app/controllers/problems_controller.rb:89:in `submit'


  Rendered /home/user/.rvm/gems/ruby-2.4.10/gems/actionpack-4.2.11.3/lib/action_dispatch/middleware/templates/rescues/_source.erb (7.4ms)
  Rendered /home/user/.rvm/gems/ruby-2.4.10/gems/actionpack-4.2.11.3/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb (6.2ms)
  Rendered /home/user/.rvm/gems/ruby-2.4.10/gems/actionpack-4.2.11.3/lib/action_dispatch/middleware/templates/rescues/_request_and_response.html.erb (2.0ms)
  Rendered /home/user/.rvm/gems/ruby-2.4.10/gems/actionpack-4.2.11.3/lib/action_dispatch/middleware/templates/rescues/diagnostics.html.erb within rescues/layout (60.6ms)
```

BelgianSalamander:
This pull request should resolve #257.

It displays an error message either when a submission's source has an invalid text encoding or the source contains a null character. Either conditions would have previously caused a 500 Internal Server Error.
- Having an invalid text encoding caused verification to throw an error when it tried checking if the source is blank (String#blank raises an exception when called on a string with an invalid encoding). This means that this needs to be checked before running any verification.
- If the source was valid UTF-8 but contained a null character, an error would be raised when trying to save the record because PostgreSQL does not support having null characters in strings.

### DIFF
--- a/app/controllers/problems_controller.rb
+++ b/app/controllers/problems_controller.rb
@@ -87,7 +87,13 @@ class ProblemsController < ApplicationController
       @problem = Problem.find(params[:id])
       @submission = Submission.new(submit_params) # create submission
       respond_to do |format|
-        if @submission.save
+        source_is_valid = @submission.source.valid_encoding? && !@submission.source.include?("\0")
+        if !source_is_valid
+          @submission.errors.add :source_file, "- Submission has an invalid text encoding. This was likely caused by submitting a compiled file (.exe, .out, .class, ...) instead of a source code file (.cpp, .c, .java, ...)."
+          @submission.source = nil; # Prevent submission form from trying to render the source (and erroring)
+        end
+
+        if source_is_valid && @submission.save
           @submission.judge
           format.html { redirect_to(@submission, :notice => 'Submission was successfully created.') }
           format.xml  { render :xml => @submission, :status => :created, :location => @submission }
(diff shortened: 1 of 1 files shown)


app/controllers/problems_controller.rb
```diff
@@ -87,7 +87,13 @@ def submit
       @problem = Problem.find(params[:id])
       @submission = Submission.new(submit_params) # create submission
       respond_to do |format|
-        if @submission.save
+        source_is_valid = @submission.source.valid_encoding? && !@submission.source.include?("\0")
```
Holmes98, line 90: Nice work. Just one slight issue – `@submission.source` seems to be `nil` if the user doesn't select a file, which causes this to raise an exception instead of displaying an error message ("Source can't be blank").

Holmes98 (approved):

PR #272, opened 2025-02-23
Allow From, Reply-To addresses to be overridden in settings table

Holmes98:
I've changed the website to send emails using train@nzoi.org.nz, but this currently causes emails to be sent with incorrect headers:

```
X-Google-Original-From: nztrain@gmail.com
Reply-To: nztrain@gmail.com
```

This would also allow setting the sender's display name, e.g. `NZOI Training <train@nzoi.org.nz>`; currently Gmail just shows the sender as "train" or "nztrain".

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -1,6 +1,4 @@
 class AdminMailer < ActionMailer::Base
-  default :from => "nztrain@gmail.com"
-
   def custom_email(admin,user,subject,msgbody)
     @user = user
     @msgbody = msgbody
--- a/app/views/user/admin_email.html.erb
+++ b/app/views/user/admin_email.html.erb
@@ -13,7 +13,7 @@
   </div>
   <p>
     <%= current_user.name %><br />
-    Admin of nztrain.com
+    Admin of NZOI Training
   </p>
 
   <div class="actions">
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,6 +47,7 @@ module NZTrain
     config.action_mailer.perform_deliveries = true
     config.action_mailer.raise_delivery_errors = true
     config.action_mailer.default :charset => "utf-8"
+    config.action_mailer.default :from => "nztrain@gmail.com"
 
     ActionMailer::Base.smtp_settings = {
       :address => "smtp.gmail.com",
--- a/config/application.rb
+++ b/config/application.rb
@@ -53,7 +53,7 @@ module NZTrain
       :address => "smtp.gmail.com",
       :port => 587,
       :authentication => :plain,
-      :domain => 'nztrain.com',
+      :domain => 'nzoi.org.nz',
       :user_name => 'nztrain@gmail.com', # username and password set with higher priority in settings table
       :password => 'training site',
       :enable_starttls_auto => true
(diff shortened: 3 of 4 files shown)


tom93:
Hi @Holmes98!
I'm curious -- why do these need to be set separately? I would expect to
have a single email address for both SMTP auth and "From". Otherwise it's
effectively spoofing the sender, which I think would be better to avoid if
possible.

Holmes98:
I don't know why, but maybe there are cases where the SMTP username isn't the same as the email address.

I guess we could still set the fields from `system/mailer/username` to avoid repetition, though then you wouldn't be able to set the display name and I didn't want to overcomplicate things.

Holmes98:
Actually, it's probably just because Action Mailer has other delivery methods than SMTP (e.g. [sendmail](https://guides.rubyonrails.org/action_mailer_basics.html#example-action-mailer-configuration)).

tom93:
Actually, how did you change the website to send emails using
***@***.***? I assumed it was by changing system/mailer/username (and
generating a password etc.), but then I don't see how there would still be
references to ***@***.***

Holmes98:
Yeah, I just changed the `system/mailer` settings, but those are only used for SMTP authentication.

For emails from Devise, the from/reply_to parameters seem to be set from the `mailer_sender` setting (but only if you don't set the default from/reply_to parameters; I still left it in though as it seems to break the specs if commented out):
https://github.com/NZOI/nztrain/blob/8e102ad7631b696d1dbd5eb62d03f9e61ba70043/config/initializers/devise.rb#L13

Emails from the admin panel currently use this (which I removed in this PR):
https://github.com/NZOI/nztrain/blob/8e102ad7631b696d1dbd5eb62d03f9e61ba70043/app/mailers/admin_mailer.rb#L2

tom93:
Right, thanks for digging that up. I consider that a bug -- we shouldn't
hard-code that email address. How about I write a separate PR to change it
to use system/mailer/username, and then this PR would only be necessary for
setting a custom display name?

Holmes98:
Sure.

tom93:
I'm mostly done (branch "pr/remove-hard-coded-nztrain-email" of my fork, [commits](https://github.com/NZOI/nztrain/compare/aec91c0c~3...aec91c0c)).

One question is how the settings should work. I would strongly prefer that in the simplest case (where the admin doesn't want to set a pretty display name), the email would only have to be configured in one setting. Here are the options I like:

**A.** "system/mailer/email" for auth and default "From:", setting "system/mailer/from" to override "From:" (may contain a display name).

**B.** "system/mailer/email" for auth and "From:", "system/mailer/display_name" (optional) added to "From:".

(For both options, a migration could rename "system/mailer/username" to "system/mailer/email".)

Also, is Reply-To useful, or was it just a workaround?

Holmes98:
That looks reasonable. I'm fine with either of those options, or even just hardcoding the display name to "NZOI Training", since there are already hardcoded references to that elsewhere on the site.

> Note that `config.action_mailer.default()` is broken[1], so we must use `default_options=`.

Nice find, I guess that's why the specs were breaking for me earlier.

Reply-To was mostly a workaround. I could see it maybe being useful for redirecting responses to an mailbox that's actually monitored, but it's probably not important.

tom93:
I implemented the hard-coded display name and opened #TD.
Are you okay with using that and dropping this PR?

PR #273, opened 2025-03-19
Set "from" email address using settings table instead of hard-coding

tom93:
Second version of #272.

This PR renames the setting system/mailer/username to system/mailer/email and uses that for both SMTP authentication and as "from" email address (previously it was used only for SMTP authentication, and the "from" address was hard-coded to "nztrain<span>@</span>gmail.com").

It also sets the sender's display name to "NZOI Training" and cleans up a couple of things.

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -1,6 +1,4 @@
 class AdminMailer < ActionMailer::Base
-  default :from => "nztrain@gmail.com"
-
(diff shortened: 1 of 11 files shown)


Holmes98 (approved):
LGTM, thanks.

tom93:
Deployed, thanks for your patience :)

screenshot-nztrain-email

(I forgot you already created #254 to use seeds.rb.)

PR #274, opened 2025-03-19
Change mailer links to use https

Holmes98:
This happens automatically with `config.force_ssl = true`, but [only from Rails 5](https://www.github.com/rails/rails/pull/17388).


tom93 (approved):
(I'll add "This happens automatically with ..." to the commit message.)

PR #278, opened 2025-04-18
Move jquery, jquery-ui, and jquery-countdown to vendored libraries

bagedevimo:
These gems aren't well maintained, don't work on the later versions of Rails (without manually reconfigured sprockets, which we could do) and generally aren't well supported as gems (most of them haven't been in updated in years). jquery-countdown & jquery in particular have dependencies on older versions of rails (pre 5.0) and updating them would also update jquery which is not a backwards compatible change.

As we don't have amazing test coverage, moving these to vendored libraries seems safer and we can continue to the same versions of these JavaScript libraries until we choose otherwise, without it being a problem with gem compatibility etc.


Holmes98 (dismissed):
I don't know if there are better solutions but it seems to work fine and I'm keen to get moving faster on upgrading Ruby/Rails.

bagedevimo:
I realised I forgot to remove jquery-countdown the gemfile, even though I had vendored the files. Woops.

tom93:
Should we switch to files downloaded from the official sources (
https://releases.jquery.com/, https://releases.jquery.com/ui/)?

vendor/assets/javascripts/jquery.js.1
tom93, line 1: Should this be deleted? It's identical to jquery.js.

tom93 (changes requested):

tom93:
There is another gem "jquery-historyjs", should we vendor that as well?

bagedevimo:
> Should we switch to files downloaded from the official sources (https://releases.jquery.com/, https://releases.jquery.com/ui/)?

Maybe? I grabbed the files in this MR from the exact gem versions that we are currently using, to minimize differences and make this release safer.

> Should this be deleted? It's identical to jquery.js.

Yes, this looks like a mistake from me. I'll fix shortly.

> There is another gem "jquery-historyjs", should we vendor that as well?

We can, but I think it's unused so I was going to suggest removing it instead. I'm just trying to be more through in checking it is unused.

tom93 (dismissed):

tom93 (approved):

Holmes98:
I'll deploy.

PR #279, opened 2025-04-18
Try to avoid running CI twice for every push to an open PR

bagedevimo:


Holmes98:
I think this will prevent CI from running if you push to a feature branch that doesn't have a PR open (which can be useful if you just want to test changes, or if you want to check that CI is passing before opening the PR).

From c5e3253:
> The template only runs the job on push/pull-request for branch
   "master", but we run the job for all branches because we want to be
   able to push commits and have them tested without having to open a
   pull request.

PR #280, opened 2025-04-19
Write out jquery.facebox.js

bagedevimo:
As far as I can tell, it's only used exactly once. This is for a modal / dialog thing that confirms your logged in user when an admin is using the "su" function to be logged in as another user.

Instead, we can redirect to another page confirm the password then redirect to wherever su would totally take us.


Holmes98 (approved):
This changes the "Cancel" button (which was previously broken anyway) to a link, but I couldn't find an easy way to make it work as a button, so I'm fine with this.

tom93:
How about

```
<input type="button" value="Cancel" style="margin-right: 20px;" onclick="history.back() || location.assign('<%= users_path %>')">
```

Holmes98:
> How about
> 
> ```
> <input type="button" value="Cancel" style="margin-right: 20px;" onclick="history.back() || location.assign('<%= users_path %>')">
> ```

Yeah, that seems better.

app/views/users/su.html.erb
```diff
@@ -4,7 +4,7 @@
     <%= password_field_tag :password, '', :autocomplete => 'off' %>
   </div>
   <div class="actions" style="padding: 0px; margin: 0px;">
-    <input type="button" value="Cancel" style="margin-right: 20px;" onClick="jQuery(document).trigger(\"close.facebox\");">
+    <%= link_to "Cancel", users_path %>
```
tom93, line 7:
```suggestion
    <input type="button" value="Cancel" style="margin-right: 20px;" onclick="history.back() || location.assign('<%= users_path %>')">
```

tom93 (approved):
(I edited the commit message and resolved merge conflicts.)

PR #284, opened 2025-04-19
Apply automatic formatting fixes from standardrb

Holmes98:
This change was made using the following commands:

```
rm .standard_todo.yml
bundle exec standardrb --fix
bundle exec standardrb --generate-todo
```


Holmes98:
> Presumably the remaining files in the to-do are those that can't be safely autocorrected?

Yeah, I might try to fix some of those in a separate PR.

app/helpers/problems_helper.rb
```diff
@@ -1,35 +1,34 @@
...
+    #   contiguous superscripts are nested, and complex values can be enclosed in parenthesis
+    #   e.g. this is the 2^(nd) time.
+    underline: true, # parse underscored emphasis as underlines. This is _underlined_ but this is still *italic*.
+    highlight: true, # parse highlights. This is ==highlighted==. It looks like this: <mark>highlighted</mark>
+    quote: true, # parse quotes. This is a "quote". It looks like this: <q>quote</q>
+    footnotes: true # parse footnotes, PHP-Markdown style.
+    #   A footnote works very much like a reference-style link:
+    #   it consists of a marker next to the text (e.g. This is a sentence.[^1])
+    #   and a footnote definition on its own line anywhere within the document
+    #   (e.g. [^1]: This is a footnote.).
```
Holmes98, line 24: Comment alignment kinda got ruined here.

Holmes98: Can ignore spacing/indentation errors here with this if we want:
```
# standard:disable Layout/CommentIndentation, Layout/ExtraSpacing
# standard:enable Layout/CommentIndentation, Layout/ExtraSpacing
```

https://github.com/standardrb/standard?tab=readme-ov-file#ignoring-multiple-lines-with-comments

tom93: I'm okay with the unaligned version.

tom93 (commented):
~~I managed to get a newer version of standardrb to work and it produces slightly different fixes, do we want to try using that?~~ *Edit: Never mind, that's not a good idea because the newer version of standardrb might use Ruby 2.5 features.*

PR #287, opened 2025-04-20
Some manual style fixes for standardrb

tom93:
Stacked on #286 (not sure if I'm doing it right, can't remember if/how I can stack PRs without using Graphite's UI).

~~(The first commit in this PR is not directly related to standardrb, but it arose out of the last commit on #286.)~~ *Changed and moved to #289 (see https://github.com/NZOI/nztrain/pull/287#pullrequestreview-2780141176)*

### DIFF
--- a/.standard_todo.yml
+++ b/.standard_todo.yml
@@ -4,7 +4,6 @@
 ignore:
 - Rakefile
 - app/controllers/contests_controller.rb
-- app/controllers/groups/members_controller.rb
 - app/controllers/importers/problem_series_controller.rb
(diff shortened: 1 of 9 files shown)


Holmes98 (approved):
Seems fine, though I don't know much about the permitted_params stuff; does the caching even matter there? From what I understand, instance variables aren't saved across different requests and I don't think we ever call `permitted_params` more than once per request.

tom93:
Good point, I'll remove the caching completely (as a new PR).

Holmes98 (approved):

PR #288, opened 2025-04-20
Update json gem to 2.7.6

tom93:
Now that we've updated to Rails 4.2 we can update json to >= 2.3 and remove the CVE patch.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby ">= 2.4.10", "< 2.5"
 
 gem "rails", "~> 4.2"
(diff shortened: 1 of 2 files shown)


tom93:
Actually, we might not need to install the json gem -- it's bundled with Ruby. I think choosing the right version would fix this.

tom93:
Never mind, the version of "json" included with Ruby 2.4 has the CVE, so we shouldn't use it.
(Also I can't find a robust way to tell bundler to use "whichever version is included with Ruby". A one-off approach is to remove "json" from Gemfile.lock, run `gem uninstall -a json`, then run `bundle install`. But that locks the version, so it won't auto-update when Ruby is updated.)

Holmes98 (approved):

PR #289, opened 2025-04-20
Remove caching from controller permitted_params etc.

tom93:
We don't call these methods multiple times during a single request, so there is no point in caching the result. (The example from the original strong_parameters gem doesn't use caching: https://github.com/rails/strong_parameters.)

This also eliminates silly inconsistencies with what was cached -- in some places it was @_permitted_params, while in other places it was the intermediate @_permitted_attributes.

Also remove method `visible_attributes` from user_controller.rb, it's unused since commit 8798951a (Refactor to remove the UserPresenter service object (#165), 2023-02-18).


Holmes98 (approved):

tom93:
Aside: Rails 8 added a new method [expect](https://api.rubyonrails.org/v8.0/classes/ActionController/Parameters.html#method-i-expect), which is recommended over require+permit ([discussion](https://blog.saeloun.com/2024/12/10/rails-8-adds-parametersexpect-to-safely-filter-and-require-params/)), but I don't think we need to try to backport & switch for now.

PR #291, opened 2025-04-23
Remove unused actions on ContestRelationsController

bagedevimo:
Most of these actions were rendered unused in 2da020ef59c5, #destroy was
restored and #update_year_level added in 00c9636d1364. The rest of these
sections seem to be dead code, so lets remove them! Git history can
always bring them back as needed, but none of the code is particularly
interesting so it'd be easy to rewrite, anyway.

I've also specs to cover off the major functions of
contest_relations_controller, including some permission tests, and
ensured that there is at least a logged in user to slightly mitigate a
leak of information in the response differing between not_found and
unathorised on ContestRelation objects.

### DIFF
--- a/spec/requests/contest_relations_request_spec.rb
+++ b/spec/requests/contest_relations_request_spec.rb
@@ -1,0 +1,81 @@
+require "spec_helper"
+
+RSpec.describe ContestRelationsController, type: :request do
+  let(:student) { FactoryBot.create(:user) }
+  let(:user) { FactoryBot.create(:superadmin) }
+  let(:contest) { FactoryBot.create(:contest) }
+  let(:school) { FactoryBot.create(:school) }
+  let!(:contest_relation) { FactoryBot.create(:contest_relation, user: student, school: school, contest: contest) }
+
+  before do
+    sign_in(user) if user
+  end
+
+  describe "DELETE /contest_relations/:id" do
+    context "when not signed in" do
+      let(:user) { nil }
+
+      context "when the relation exists" do
+        it "doesn't remove the relation" do
+          expect { delete "/contest_relations/#{contest_relation.id}" }
+            .not_to change { contest.reload.contest_relations.count }
+        end
+
+        it "redirects to login page" do
+          delete "/contest_relations/#{contest_relation.id}"
+          expect(response).to redirect_to(new_user_session_path)
+        end
+      end
+
+      context "when the relation doesn't exist" do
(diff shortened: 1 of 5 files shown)


spec/requests/contest_relations_request_spec.rb
```diff
@@ -0,0 +1,81 @@
+require "spec_helper"
+
+RSpec.describe ContestRelationsController, type: :request do
+  let(:student) { FactoryBot.create(:user) }
+  let(:admin) { FactoryBot.create(:superadmin) }
```
Holmes98, line 5: I think I understand how this works now, but I was a bit confused at first since `admin` isn't always an admin. Might be clearer if renamed to `user` or `current_user`?

spec/requests/contest_relations_request_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+            .not_to change { contest.reload.contest_relations.count }
+        end
+
+        it "redirects to login page" do
+          delete "/contest_relations/#{contest_relation.id}"
+          expect(response).to redirect_to(new_user_session_path)
+        end
+      end
+
+      context "when the contest doesn't exist" do
```
Holmes98, line 30:
```suggestion
      context "when the relation doesn't exist" do
```

spec/requests/contest_relations_request_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      context "when the relation exists" do
+        it "removes the relation" do
+          expect { delete "/contest_relations/#{contest_relation.id}" }
+            .to change { contest.reload.contest_relations.count }
+            .from(1).to(0)
+        end
+      end
+
+      context "when the contest doesn't exist" do
```
Holmes98, line 51:
```suggestion
      context "when the relation doesn't exist" do
```

Holmes98 (commented):
I think this is mostly reasonable. Commit message has a few typos ("I've also specs", "unathorised"). Also note that #destroy was restored in 112b236.

bagedevimo:
Nice catches @Holmes98! I've setup an old laptop for deving on this project, and haven't got some of my tools working yet (like vim's spellchecker!). Fixed those other things you've spotted, and amended the commit message to include that correct commit ref. You'll notice the sentence has proper grammar now, as I always ended to put that commit sha there but totally forgot.

Holmes98 (approved):

PR #292, opened 2025-04-24
Remove most XML responses

bagedevimo:
Almost none of these were being used usefully (as checked by a quick
grep of the logs on the server) except for SubmissionController#show,
which I have left in place for now. I also removed all comments refering
to routes, as a great number of them are incorrect and we discussed
briefly on discord and decided to remove. (use `bundle exec rake routes`
instead. Additionally, a fixed a couple of cases of record#destory being
called instead of record#destroy!. You pretty much always want to use
destroy! - the former version has no errors or warnings if the delete
fails for any reason, which is rarely what you want.


bagedevimo:
* **#292** <a href="https://app.graphite.dev/github/pr/NZOI/nztrain/292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/NZOI/nztrain/292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

app/controllers/submissions_controller.rb
```diff
@@ -24,40 +22,27 @@ def index
...
     @submission = Submission.find(params[:id])
     authorize @submission, :show?
-    respond_to do |format|
-      format.html # show.html.erb
-      format.xml { render xml: @submission }
+
+    if params.key?(:debug) && current_user.is_admin?
+      render xml: @submission
+    else
+      render :show
```
Holmes98, line 34: If we're keeping this, I'd prefer keeping the `.xml` syntax if possible rather than changing to `?debug`, since it's clearer about the response format and makes it easier for other formats to be added in the future (e.g. JSON). Potentially either of these?

```suggestion
    respond_to do |format|
      format.html # show.html.erb
      format.xml { render xml: @submission } if current_user.is_admin?
    end
```

```suggestion
    if request.format.xml? && current_user.is_admin?
      render xml: @submission
    else
      render :show
```

Otherwise I don't mind if we just remove the XML response from here as well for now (I think the cases you found in the logs were probably just me testing that from when you originally asked about this).

Holmes98 (approved):

bagedevimo:
### Merge activity

* **May 18, 9:07 AM EDT**: @bagedevimo merged this pull request with [Graphite](https://app.graphite.dev/github/pr/NZOI/nztrain/292).

PR #302, opened 2025-06-09
Configure the hybrid cookie session store

bagedevimo:
Rails 4.1 introduced the new JSON format, and it becomes default in
Rails 5.0. We want to be using hybrid in production for a period of time
in order to migrate everyone to JSON before letting JSON be the default.

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#cookies-serializer

### DIFF
--- a/config/initializers/cookies_serializer.rb
+++ b/config/initializers/cookies_serializer.rb
@@ -1,0 +1,5 @@
+# Be sure to restart your server when you modify this file.
+
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid


config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# Be sure to restart your server when you modify this file.
+
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :marshal
```
Holmes98, line 5: Was this supposed to be `:hybrid`?

bagedevimo: No, I need to update the commit message I think instead. I wanted to add this file before the rails-5 change, then we can switch to hybrid later. I better go find the docs / changelog / recommendations again and check what versions have what options available. Please hold!

bagedevimo: Okay, I found better documentation, and I have updated the documentation to include links to that docs. We can (yay!) use hybrid now.

Holmes98 (approved):

bagedevimo:
### Merge activity

* **Jun 11, 7:24 AM UTC**: @bagedevimo merged this pull request with [Graphite](https://app.graphite.dev/github/pr/NZOI/nztrain/302).

PR #307, opened 2025-06-30
Fix clipping in expanded problem set item lists

Holmes98:
Alternative fix to #305/#306 that avoids adding a scrollbar and preserves the linear transition.


puqeko (approved):
Verified for dashboard and groups. Happy with this solution.

PR #310, opened 2025-07-03
Remove JS and CSS asset compressors

tom93:
TO DO:
- [ ] Test (update: did light testing, no issues)
- [ ] Compare size of assets before/after

Stacked on #308 and #309.

From the commit message:

> Rails 8 has moved away from asset compression[1].
> 
> `uglifier` (JS compressor) requires a JavaScript runtime (which we
> previously provided using mini_racer). It's a large install and raises
> security concerns because it uses an outdated version of V8 (it's only
> meant to run as part of asset precompilation, not in the running web
> server; but still).
> 
> `yui` requires the Java Runtime Environment.
> 
> We follow the Rails 8 rationale and remove the asset compression (even
> though we are still on a much older version of Rails).
> 
> [1] https://guides.rubyonrails.org/v8.0/asset_pipeline.html

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -86,17 +86,3 @@ group :development, :test do
 
   gem "rails-controller-testing", "~> 1.0"
 end
-
-# Gems used only for assets and not required
-# in production environments by default.
-# group :assets do
-gem "uglifier", ">=1.0.3"
-# Provide a JS runtime to execjs without needing to
-# have node, bun, or similar installed on the relevant
-# server. We should write this out asap, along with the
-# rest of the gems in this assets category, as bundling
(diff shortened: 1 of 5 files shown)


tom93:
Definitely.

I created this PR because I got annoyed with `bundle install` running out of RAM while testing Rails 5 (a VM with 1GB isn't enough, need to enable swap). No rush to merge, just created it so I won't forget.

I assumed Uglifier was only compressing (not adding ES compat) and I was hoping everything would "just work" (but as I said I didn't even bother testing), do you have any particular concerns / things to test? I remember talking about importmaps for auto-updating jQuery, but I'm not sure why else we would need importmaps / non-trivial asset compilation.

PR #312, opened 2025-07-03
Refactor Request.closed

tom93:
@bagedevimo I saw your "TODO(Rails5)" and tried to make the change but got errors; I'm creating this PR as a reminder (feel free to close if it's not useful).


Holmes98:
https://stackoverflow.com/questions/7054188/is-it-possible-to-negate-a-scope-in-rails

Seems that you can't negate a scope with `where.not`, so might have to be something like this:
```ruby
  scope :status_not_pending, -> { where.not(status: STATUS[:pending]) }
  scope :closed, -> { status_not_pending.or(past_expiry) }
```
(also I think the comment should've used `scope_pending` rather than `pending`)

PR #313, opened 2025-07-03
Remove superfluous return value from 'before' callbacks

tom93:
Reminder for myself to look into this (help welcome). I thought it would be a simple but I get errors; maybe this needs to wait for Rails 5.1, or maybe I misunderstood the [documentation](https://guides.rubyonrails.org/v5.2/upgrading_ruby_on_rails.html#halting-callback-chains-via-throw-abort), specifically what they mean by "return". Relevant Rails code: [deprecation](https://www.github.com/rails/rails/pull/17227), [removal](https://www.github.com/rails/rails/commit/3a25cdca3e0d29ee2040931d0cb6c275d612dffe).


Holmes98:
> I think I tried the configuration to get the new behavior and it didn't help

If you were setting it in config/application.rb, it's being overridden by this:
https://github.com/NZOI/nztrain/blob/5f00a74f80b986ded0f42e39c9bb8e077ab6dde7/config/initializers/new_framework_defaults.rb#L24-L25

I'll push some changes here to opt into the new behavior, hopefully that's fine.

bagedevimo (approved):

PR #314, opened 2025-07-04
Remove config option raise_in_transaction_callbacks

bagedevimo:
This is now default in Rails 5.0, so we no longer need this opt-in
config option.

https://guides.rubyonrails.org/v6.1.0/5_0_release_notes.html#active-record-notable-changes


Holmes98:
Just for reference, this was added in #220.

PR #315, opened 2025-07-04
Remove some unused(?!) methods on contest_relation

bagedevimo:
alias_method_chain is deprecated in Rails 5.0 and removed in Rails 5.1,
so we need to stop using it. I can't find any way these are being
called, unless it ends up being ActiveRecord implict nonsense I can't
find evidence of.

### DIFF
--- a/app/models/contest_relation.rb
+++ b/app/models/contest_relation.rb
@@ -67,16 +67,11 @@ class ContestRelation < ActiveRecord::Base
     update_finish_at
   end
 
-  def contest_with_update=(contest)
-    self.contest_without_update = (contest)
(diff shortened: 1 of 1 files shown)


Holmes98:
`alias_method_chain :contest=, :update` basically seems to be an shorthand for
```ruby
alias_method :contest_without_update=, :contest=
alias_method :contest=, :contest_with_update=
```
So it was being used to modify the behaviour of `contest=`, while preserving access to the original method through `contest_without_update=`. But I think since we aren't using `contest_without_update=`, we can just directly override `contest=` instead? I've pushed a potential fix in a9ea1d7.

PR #316, opened 2025-07-04
Remove serve_static_files and enable public_file_server from config

bagedevimo:
config.public_file_server was added in like Rails 3, and
serve_static_files is deprecated in 5.0 and removed in 5.1.

The relocations are to have these files match up more with the default
generated files by Rails.


tom93:
Similar thing re-done in #324.

PR #317, opened 2025-07-04
Replace kwargs request specs with params objects

bagedevimo:
Deprecated in Rails 5, removed in 5.1.


tom93:
How does this compare to my commit in rails-5-suggestions-v2? (90ab1e64,
Fix deprecation warning in specs (get/post positional args))

Nit: parans => params, and standardrb doesn't like spaces inside {}.

tom93:
Re-done in #324.

PR #323, opened 2025-07-05
Reduce DB queries for schools when using scoreboard

### LINKED ISSUE #151: Reduce DB queries for schools when using scoreboard [small]
https://github.com/NZOI/nztrain/blob/743679784bebc5bc2a37e7e10d83bef6d033395d/app/models/contest.rb#L75

> i think if you do a .includes(user: :school) you'll skip an N+1 accessing schools

--@bagedevimo in https://github.com/NZOI/nztrain/pull/142#discussion_r1098525089 (edited)

Need to confirm that we are making extra DB queries (in production mode) on the schools table when displaying / exporting the scoreboard, and that the proposal fixes that.

Holmes98:
Fixes #151.

### DIFF
--- a/app/controllers/contests_controller.rb
+++ b/app/controllers/contests_controller.rb
@@ -90,7 +90,7 @@ class ContestsController < ApplicationController
     @contest = Contest.find(params[:id])
     authorize @contest, :contestants?
     @groups = Group.all
-    @contest_relations = @contest.contest_relations.includes(:user).order("contest_relations.school_id, users.username").references(:user)
+    @contest_relations = @contest.contest_relations.includes(:user, :school).order("contest_relations.school_id, users.username").references(:user)
 
     render layout: "contest"
   end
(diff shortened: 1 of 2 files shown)


tom93:
I'm trying to verify this, but I get few queries even without this patch. So merging this shouldn't hurt, but for the benefit of my curiosity, do you know what's going on?
(I'm using RAILS_ENV=production, it mostly works -- just need to add `confirmed_at: Time.current` to db/seeds/common.rb and set `config.force_ssl = false` in production.rb, I also set `config.log_level = :debug`. I can share my scripts for creating a fake contest with lots of users, I made all of them have the same school.)

Holmes98:
Hmm, it might require having the "Only rank official contestants" setting enabled? (due to the condition in https://github.com/NZOI/nztrain/blob/a87a89ac91471e3fddc8403aa2bd87416699fccf/app/views/contests/scoreboard.html.erb#L53C1-L53C108)

Holmes98:
Ah yeah, I think that can be fixed with a similar change (adding `includes(:school)` to `ContestsController#contestants` in contests_controller.rb

Holmes98:
I've just added that here, could be squashed.

tom93 (approved):
Verified both pages!

PR #324, opened 2025-07-05
Fix several Rails 5.0 deprecations

tom93:


Holmes98 (approved):
Seems fine to me. Presumably the spurious deprecation warnings about positional args would go away after we upgrade to Rails 5.1?

tom93:
Yes (I didn't test to confirm this, but the Rails source code seems clear). I rewrote the commit message to reflect that. (Look ma, no review invalidation!)

PR #326, opened 2025-07-05
Update Bundler install script to use Bundler 1.17

tom93:
Gemfile.lock recently changed to say "BUNDLED WITH 1.17.3", so now Bundler 1.16 gives a warning.


tom93:
I'll do one more round of tests on my recent PR's and then I'll do a merge spree, thanks for approving. Will manually update Bundler on the server.

PR #334, opened 2025-07-18
Update Python

puqeko:
* Adds Python 3.14 as default.
* Adds PyPy 3.11 as other.

CPython is compiled from source while binaries are fetched for PyPy.


Holmes98:
I'm not strongly opposed to this, but I'd prefer that we try to upgrade/replace the server off of xenial first, as that would make the Python upgrades much simpler and otherwise it kinda gives us an excuse to stay on xenial for longer.

Also slightly hesitant about using the pre-release version in production; probably safer to use 3.13 or to wait for the stable release of 3.14.

PR #340, opened 2025-07-20
Add puma to Gemfile

bagedevimo:
We can probably switch from nginx-passenger to nginx + puma in
production, which means a few good things!

1. We'd be back to the rails default (since 5.0)
2. We'd run the same web server in dev and in prod (currently thin vs
   passenger)
3. We get the easy of maintenance and running of puma in production (vs
   passenger requiring all sorts of socket nonsense, etc)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -99,3 +99,5 @@ gem "uglifier", ">=1.0.3"
 gem "mini_racer", "~> 0.4.0"
 gem "yui-compressor"
 # end
+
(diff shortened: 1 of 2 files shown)


Holmes98 (approved):

bagedevimo:
### Merge activity

* **Jul 20, 7:20 AM UTC**: @bagedevimo merged this pull request with [Graphite](https://app.graphite.dev/github/pr/NZOI/nztrain/340).

PR #348, opened 2026-01-11
Fix rounding issues that appeared with subtask scoring

BelgianSalamander:
This fix should remove weird rounding issues that cropped up with subtask scoring.

For example, a score that should be 58 would sometimes be displayed as 57 since due to floating point imprecision it was stored as 57.99999999 before getting rounded down.

This pull request should remove such issues while minimising any other visible changes (so most scores wouldn't change at all)


Holmes98 (approved):
A bit hacky, but it should be fine for now.

PR #349, opened 2026-06-21
Return database connections to the pool after each qless job

Holmes98:
This fixes an issue where workers could fail to judge submissions if the database connection was dropped (e.g. because postgresql was restarted).

https://discord.com/channels/670126531489824788/678154623571329025/1485894172748550185

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -27,6 +27,18 @@ namespace :qless do
 
     Qless::Workers::ForkingWorker.send(:include, ActiveRecordReconnect)
 
+    module ActiveRecordReconnectJob
+      def around_perform(job)
+        ActiveRecord::Base.connection_pool.with_connection do
+          super(job)
+        end
+      end
+    end
+
+    Qless::Workers::BaseWorker.class_eval do
+      include ActiveRecordReconnectJob
+    end
+
     # The only required option is QUEUES; the
     # rest have reasonable defaults.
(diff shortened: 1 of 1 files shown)


tom93:
A summary of some messages from Discord for posterity:

The qless README suggests using ActiveRecord::Base.establish_connection, but that caused a PG::UnableToSend exception (hence the use of with_connection in this PR).
There is a similar stack trace for establish_connection at https://www.github.com/rails/rails/issues/12867#issuecomment-28690381, and it looks like the proper fix for that is https://www.github.com/rails/rails/pull/29785 but that's only in Rails 5.2.
Another suggestion from related issues is to tweak the TCP keepalive: https://www.github.com/rails/rails/issues/20411#issuecomment-360801422 (or https://www.github.com/rails/rails/issues/9421#issuecomment-41952270), might be worth trying if there are still problems in the future.

tom93 (approved):
I didn't test this, but it seems safe enough (even without getting to the bottom of why it works while establish_connection fails).
*Edit:* Not sure if this will work if the database is restarted while a submission is being judged...

Holmes98:
> Not sure if this will work if the database is restarted while a submission is being judged

Yeah, it'll still throw an error in that case, but this at least seems to allow workers to automatically recover from the broken connection (on the next job) rather than failing on all subsequent submissions. Hopefully it's good enough for now.

Thanks for reviewing.

